PR #2, opened 2016-09-14
added new possible fields for the users.identity method

jasonmullins:


pom.xml
```diff
@@ -5,7 +5,7 @@
 
     <groupId>com.github.seratch</groupId>
     <artifactId>jslack</artifactId>
-    <version>1.0.1</version>
+    <version>1.0.2</version>
```
seratch: You don't need modify the version. I'll do that when releasing it.

src/main/java/com/github/seratch/jslack/api/methods/response/users/UsersIdentityResponse.java
```diff
@@ -17,10 +18,24 @@
     public static class User {
         private String name;
         private String id;
+        private String email;
+        @SerializedName("image_24")
```
seratch: Ah, is this needed?

seratch:
This pull request refers https://github.com/seratch/jslack/issues/1 but is it still WIP?

pom.xml
jasonmullins: I reverted it.

src/main/java/com/github/seratch/jslack/api/methods/response/users/UsersIdentityResponse.java
jasonmullins: Yes - I tried it without it but it didn't work.  It would probably work if I just named the fields like "image_24" instead of "image24" but I wanted to be consistent with com.github.seratch.jslack.api.model.User.

jasonmullins:
I think I included all the data that is documented here: https://api.slack.com/methods/users.identity.

This includes scopes: "identity.basic", "identity.email", "identity.avatar", and "identity.team".  I think that's all for the "identity" scope unless I missed something.

PR #13, opened 2016-12-23
oauth_bot added possible data on OAuthAccessResponse

jasonmullins:
I added missing data to OAuthAccessResponse


seratch:
Thanks 👍

PR #26, opened 2017-05-29
bug/post-message-with-thread-ts fixed bug for post message in the eve…

jasonmullins:
…nts api


seratch:
@jasonmullins Thank you for your contribution! I just published version 1.0.11 which includes your fix.

PR #28, opened 2017-07-13
'ts' -- should be a String

c0mpiler:
'ts' values are more or less like -->  1498530147.464010
which is not an integer. So changing it to String instead of Integer --> keeping it consistent with the 'ts' attribute in the 'message' model


seratch:
LGTM

seratch:
Thanks, I just published version 1.0.12 to the sonatype oss repositories. It will be synched to maven central within a few hours.

c0mpiler:
Thank you very much for the quick response. Very much appreciated!!

Issue #20, opened 2017-04-18
is support for message menus planned?


n0mer:
https://slackhq.com/take-your-pick-75767e13b497

seratch:
That sounds good. 👍  https://api.slack.com/docs/message-menus
Recently, I myself am a bit busy. If somebody implements the enhancement, I'd love to merge it.

PR #29, opened 2017-10-06
Add support for message menus

joeltoby:
- Fix for issue #20 
- Added a new attachment option Enum for easy switch between button and menu
option types.


seratch:
Thanks. Although your pull request brings an API breaking change, it's good for me to merge this because enhancing the `type` looks reasonable.

joeltoby:
Apologies @seratch.  I knew it was a breaking change and forgot to mention it.

I figured that as `type` button was the default anyway, most clients wouldn't have explicitly declared the type and so wouldn't experience a breakage.  I should have mentioned it though!

Issue #31, opened 2017-10-13
Add support for Slack dialog modals


joeltoby:
Dialogs were recently released by Slack (see [here](https://api.slack.com/dialogs)) and a `dialog.open` web API has also been exposed.

Request to add support to this API in jslack.

PR #33, opened 2017-10-13
Add support for modal dialogs

joeltoby:
Dialogs were recently released by Slack (see here) and a dialog.open web API has also been exposed (see issue #31).

Added support for dialogs including new models
Added sample usage to the README.md

**Note: No test added for this API**. dialog.open requests can only be made with a triggerID received within the last 3 seconds, and generating a triggerID requires a user taking action such as clicking a button. Testing will be more challenging!


seratch:
I have two requests on your pull request:
- write a simple test to run the code
- use consistent indent (4 spaces)

joeltoby:
Thanks for reviewing it @seratch.

How would you like the test to work?  The simplest way I can think of would be to send a message button to a channel / DM which a user would have to click.  Once the button is clicked, a callback handler would then respond with the dialog (within the required 3 seconds).  The user action of course means that the testing cannot be fully automated.

Alternatively, what are your thoughts on integrating with [Steno](https://slackapi.github.io/steno/) (a Slack maintained testing framework for the SlackAPI)?  My understanding is that it can be used to mock API requests.

README.md
```diff
@@ -223,6 +224,56 @@ ChatDeleteResponse deleteResponse = slack.methods().chatDelete(
 assertThat(deleteResponse.isOk(), is(true));
 ```
 
+##### Open a dialog modal
+
+```java
+String token = "api-token";
+    
+    // Required.  See https://api.slack.com/dialogs#implementation
```
seratch: You don't need to have indent from this line.

seratch:
>How would you like the test to work? The simplest way I can think of would be to send a message button to a channel / DM which a user would have to click. Once the button is clicked, a callback handler would then respond with the dialog (within the required 3 seconds). The user action of course means that the testing cannot be fully automated.

I see. I just want some tests to easily run the method. It's okay even if we need to close the dialog manually.

joeltoby:
> I see. I just want some tests to easily run the method. It's okay even if we need to close the dialog manually.

The issue is more the opening of the dialog than the closing.  A dialog in Slack can only be opened in response to a user action such as a slash command or button click.  Slack recently added `trigger_id`s callback responses, and a `dialog.open` request has to include that same `trigger_id` in order to succeed.  The `dialog.open` request must also be made within 3 seconds of the user action.

README.md
joeltoby: Done.  Commit: https://github.com/seratch/jslack/pull/33/commits/823f7197a57defaae3cd4d12472a0f5321e05aa4

joeltoby:
> use consistent indent (4 spaces)

Done. 
 See commit https://github.com/seratch/jslack/pull/33/commits/365f91a3614d5c0ba32acf8c8741736afe31d85d

src/main/java/com/github/seratch/jslack/api/model/dialog/Option.java
```diff
@@ -0,0 +1,15 @@
...
+import lombok.Builder;
+import lombok.Data;
+
+/**
+ * A dialog element option used with {@link DialogSelectElement}s
+ *
+ */
+@Data
+@Builder
+public class Option {
```
seratch: We already have a generic `Option` class. https://github.com/seratch/jslack/blob/master/src/main/java/com/github/seratch/jslack/api/model/Option.java

`DialogOption` is preferred for me. Is it fine with you to rename this?

Also, I would like to have `Dialog` prefix in all classes under `com.github.seratch.jslack.api.model.dialog` for consistent naming (= Element -> DialogElement, SubType -> DialogSubType).

src/test/java/com/github/seratch/jslack/Slack_dialogs_Test.java
```diff
@@ -0,0 +1,56 @@
...
+        Dialog dialog = Dialog.builder()
+        		.title("Request pens")
+        		.callbackId("pens-1122")
+        		.elements(Arrays.asList(quanityTextElement))
+        		.submitLabel("")
+        		.build();
+        
+        DialogOpenResponse dialogOpenResponse = slack.methods().dialogOpen(DialogOpenRequest.builder()
+        		.token(token)
+        		.triggerId("FAKE_TRIGGER_ID")
```
seratch: OK, so far. Could you have a short comment when we cannot a valid trigger id here as you mentioned in the description of this pull request?

src/main/java/com/github/seratch/jslack/api/model/dialog/Option.java
joeltoby: Yes- that will be clearer. I'll rename the classes

src/test/java/com/github/seratch/jslack/Slack_dialogs_Test.java
joeltoby: No problem.

What do you think about mocking tests in a future version using either Slack's Steno project or mockito? Will I open an issue to discuss this?

seratch: >What do you think about mocking tests in a future version using either Slack's Steno project or mockito?
It's very cool. If you have time to work on it, please introduce mocking tests for this.

joeltoby: >It's very cool. If you have time to work on it, please introduce mocking tests for this.

Thanks.  I'll do that in a separate PR.

joeltoby:
@seratch Apologies for this very messy PR!

I've made all the changes requested now.

seratch:
LGTM, thanks.

PR #40, opened 2017-11-04
Resolves #38

### LINKED ISSUE #38: Allow configuring URL prefixes
First, thanks for your sharing jslack 👍 

I'm using the Slack Methods client and I noticed the API prefix is hardcoded in a `static final` variable in `MethodsClientImpl#ENDPOINT_URL_PREFIX`. It'd be very useful if we could override those prefixes from outside.

A common scenario is when you want to use something like [wiremock](http://wiremock.org/) for writing integration tests using mock calls. In those cases you need to hit the stub server instead of the real one, so you need to configure a different URL prefix. You'll need this for #36 too. I could send you a PR if you don't mind.

cesdperez:


src/main/java/com/github/seratch/jslack/api/methods/impl/MethodsClientImpl.java
```diff
@@ -111,14 +111,19 @@
...
+    private String endpointUrlPrefix = "https://slack.com/api/";
 
     private final SlackHttpClient slackHttpClient;
 
     public MethodsClientImpl(SlackHttpClient slackHttpClient) {
         this.slackHttpClient = slackHttpClient;
     }
 
+    @Override
+    public void setEndpointUrlPrefix(String endpointUrlPrefix) {
```
joeltoby: Would a `getEndpointUrlPrefix` method also be useful?

joeltoby:
This will be really useful. Thanks @cesdperez

PR #45, opened 2017-11-09
Add new fields to message object & groupReplies function

scoobymib3:
Added username, botId, subtype, file and icons to Message object to allow more detailed display


scoobymib3:
Apologies, these two should have been separate requests. First pull request so was always going to make a mistake somewhere :)

seratch:
Thanks.

PR #46, opened 2017-11-14
[issue-39] reuse httpclient between invocations

esiqveland:
This change makes the Slack object reuse it's SlackHttpClient.
Notice that even if Slack object is provided a SlackHttpClient in its builder function,
this instance was never used:
https://github.com/seratch/jslack/blob/master/src/main/java/com/github/seratch/jslack/Slack.java#L42

Recreating the underlying OkHttpClient is a very expensive operation,
setting up threadpools and re-establishing TCP connections/https handshakes,
and should be avoided.


esiqveland:
Many tests are failing, but it looks to be because the SLACK AUTH token set, is no longer valid.

seratch:
I confirmed all tests passed on my local machine. Thanks for working on this.

PR #53, opened 2018-01-04
Fix typo: should use is_mpim instead of is_npim

ramzimaalej:
According to the doc: https://api.slack.com/types/group
  
  cc @seratch


seratch:
Thanks!

seratch:
Thanks, I've released version 1.0.20 which contains this change.

ramzimaalej:
thanks a lot!

PR #56, opened 2018-01-24
Log errors, provide interfaces for custom close and error handlers

farmdawgnation:
Hi there!

While using jslack for a Slack bot, we noticed that the bot will periodically just "fall over" with no indication of why in the logs. Messages just stop arriving and it stops responding. Taking a look at the implementation of `RTMClient` we had relatively few hooks to determine what was going on. We also noticed that `RTMClient` didn't attach an `OnError` handler.

This PR:

* Adds an `OnError` handler to the `RTMClient`
* Provides for a `RTMErrorHandler` and `RTMCloseHandler` to be provided so that applications that depend on jslack have a clean interface to attach custom logic into the close / error pipelines.

Please let me know if there are any questions or concerns with this approach. :)


seratch:
Thank you for your contribution. I just released 1.0.21 which contains your changes. https://github.com/seratch/jslack/compare/1.0.20...1.0.21

PR #57, opened 2018-01-29
Add URL field to support link buttons

lewisheadden:
[Link Buttons](https://api.slack.com/docs/message-attachments#link_buttons) require that `Action`s can have the URL field set on them.


seratch:
Thanks, I just released 1.0.22 which contains your change.

PR #58, opened 2018-01-30
Adding error metadata to DialogOpenResponse

louis-pre:
Hello! 
While trying to work with the dialog.open method, I encountered formatting issues. Slack reports them in the response they return but it was not accessible through the DialogOpenResponse object.
(see https://api.slack.com/methods/dialog.open, response_metadata)

I have added them to the dialog response and named the object ErrorMetadata because ResponseMetadata already exists (for chat responses, as I understand it).

I hope it does not break any of your conventions :)

### DIFF
--- a/src/main/java/com/github/seratch/jslack/api/model/ErrorMetadata.java
+++ b/src/main/java/com/github/seratch/jslack/api/model/ErrorMetadata.java
@@ -1,0 +1,12 @@
+package com.github.seratch.jslack.api.model;
+
+import lombok.Data;
+
+import java.util.List;
+
+@Data
+public class ErrorMetadata {
+
+    private List<String> messages;
+
+}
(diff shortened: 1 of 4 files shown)


src/main/java/com/github/seratch/jslack/api/model/ErrorMetadata.java
```diff
@@ -0,0 +1,12 @@
+package com.github.seratch.jslack.api.model;
+
+import lombok.Data;
+
+import java.util.List;
+
+@Data
+public class ErrorMetadata {
```
seratch: How about renaming this to `com.github.seratch.jslack.api.model.dialog.DialogResponseMetadata`?

louis-pre:
Done! 
Can you do a Maven deployment when merge?

Thanks a lot :)

src/main/java/com/github/seratch/jslack/api/model/ErrorMetadata.java
seratch: Please change its package as well.

seratch: Please change its package as well.

seratch:
I wil do. After you change the package to place the class, I will merge this PR.

louis-pre:
Thanks, sorry about that.

seratch:
Thank you for your contribution. I've released 1.0.23 which contains your contribution :tada:

PR #77, opened 2018-07-24
Make subclassing MethodsClientImpl easier

### LINKED ISSUE #75: MethodsClientImpl is difficult to subclass
[`MethodsClientImpl`](https://github.com/seratch/jslack/blob/master/src/main/java/com/github/seratch/jslack/api/methods/impl/MethodsClientImpl.java) implements over 130 API methods from the [`MethodsClient`](https://github.com/seratch/jslack/blob/master/src/main/java/com/github/seratch/jslack/api/methods/MethodsClient.java) interface.  Each of these methods in turn invokes one of three **private** `doPost...` methods.

I would like to extend `MethodsClientImpl` within my project to add some custom rate limiting, but this would require overriding each of the 130+ methods (as the private `doPost...` methods cannot be overridden).

Changing the access modifiers on the three `doPost` methods from **`private` to `protected`** would make it much easier for clients to write custom MethodsClient implementations.

joeltoby:
Changes the access modifiers on MethodsClientImpl's three private doPost methods to protected This makes it much easier to extend. The issue is fully described here #75.

Fixes #75

### DIFF
--- a/src/main/java/com/github/seratch/jslack/api/methods/impl/MethodsClientImpl.java
+++ b/src/main/java/com/github/seratch/jslack/api/methods/impl/MethodsClientImpl.java
@@ -1421,7 +1421,7 @@ public class MethodsClientImpl implements MethodsClient {
         }
     }
 
-    private <T> T doPostForm(
+    protected <T> T doPostForm(
             FormBody.Builder form,
             String endpoint,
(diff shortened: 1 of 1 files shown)


joeltoby:
@seratch - Sorry for opening and closing my PR.  I had a GIT issue which is resolved now

PR #80, opened 2018-10-15
fix some build warnings

SylvainJuge:
- remove default plugin **groupId** for default plugins
- set fixed plugin versions for more stable build, otherwise using "current plugin version" may break over time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,4 +5,4 @@ jdk:
   - oraclejdk9
   - oraclejdk10
 install: echo 'skipped'
-script: ./mvnw test-compile
+script: ./mvnw test
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,4 +5,4 @@ jdk:
   - oraclejdk9
   - oraclejdk10
 install: echo 'skipped'
-script: ./mvnw test
+script: ./mvnw test-compile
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -5,4 +5,4 @@ jdk:
   - oraclejdk9
   - oraclejdk10
 install: echo 'skipped'
-script: ./mvnw test-compile
+script: ./mvnw test
```
seratch: Somehow, the script returns 0. But tests don't pass without an access token.
```
Tests run: 52, Failures: 29, Errors: 15, Skipped: 2
[ERROR] There are test failures.
```

SylvainJuge: I didn't knew that, is that what is mentioned in this issue https://github.com/seratch/jslack/issues/36 ?

Actually it was also a convenient way for me to start build again because it failed downloading maven at first execution.

seratch: Anyway, I cannot merge this change. Except for this one, I'd love to merge your improvements. Please revert only this change + squash commits into one commit.

SylvainJuge:
I've just reverted the last commit, you can merge it then !

seratch:
Thank you very much.

PR #81, opened 2018-10-17
re-enable latest jdk build

SylvainJuge:
I had to disable jdk10 build to make my previous PR work on travis (build is not really reliable).
This re-enables jdk10. I may try to add jdk11 if jdk10 works

oracle jdk10 seems broken, whereas 8,9 and 11 work as expected.
should we temporary replace oracle jdk 10 with openjdk 10 then ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ jdk:
   - oraclejdk8
   - oraclejdk9
   - oraclejdk10
+  - oraclejdk11
 cache:
   directories:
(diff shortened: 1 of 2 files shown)


seratch:
To be honest, supporting OracleJDK 10 is not so important because its lifetime is quite short. It's okay to replace it with openjdk10 or just remove it. I'm fine if we can support oraclejdk8 and openjdk11 at least because they're long time supported JDKs.

PR #98, opened 2019-03-09
Upgrade oss-parent to 9

### LINKED ISSUE #97: Upgrade oss-parent to 9
The latest version of the oss-parent POM is 9. jslack currently uses version 7.

I'll craft a PR for this, but need to reference an issue nr in it.

dashorst:
Fixes #97


seratch:
Thanks :)

PR #99, opened 2019-03-09
Adds Block Kit support

### LINKED ISSUE #96: Support Block Kit
https://api.slack.com/block-kit

dashorst:
This is my first code contribution to jslack. I expect that my first attempt
probably has some issues and that my choices may not be yours. You are
welcome to provide feedback, ask for modifications or modify the code.

I submit this patch under the same license as jslack (currently MIT) and
transfer ownership of the code to @seratch. 

The description of this pull request:

Slack has created a new (improved?) message format based on layout
blocks that allow for semi-advanced rich formatting called Block Kit:

https://api.slack.com/block-kit

This commit adds support for all elements, while retaining API
compatibility with the old API (which is discouraged by the Slack
documentation https://api.slack.com/messaging/composing/layouts#secondary-attachments)

The new BlockMessagePayload builder allows you to start crafting your
message using layout blocks.

The canonical example from the Block Kit Builder, built with the jslack
model:

```
BlockMessagePayload payload = BlockMessagePayload.builder()
	.blocks(Arrays.asList(
		SectionBlock.builder()
			.text(MarkdownTextObject.builder()
				.text("Hello, Assistant to the Regional Manager Dwight! *Michael Scott* wants to know where you'd like to take the Paper Company investors to dinner tonight.\n\n *Please select a restaurant:*")
				.build())
			.build(),
		DividerBlock.builder().build(),
		SectionBlock.builder()
			.text(MarkdownTextObject.builder()
				.text("*Farmhouse Thai Cuisine*\n:star::star::star::star: 1528 reviews\n They do have some vegan options, like the roti and curry, plus they have a ton of salad stuff and noodles can be ordered without meat!! They have something for everyone here")
				.build())
			.accessory(ImageElement.builder()
				.imageUrl("https://s3-media3.fl.yelpcdn.com/bphoto/c7ed05m9lC2EmA3Aruue7A/o.jpg")
				.altText("alt text for image")
				.build())
			.build()
		))
	.channel("#test")
	.threadTs("1552173794.000500")
	.build();
```

Fixes #96

### DIFF
--- a/src/main/java/com/github/seratch/jslack/api/model/block/ActionsBlock.java
+++ b/src/main/java/com/github/seratch/jslack/api/model/block/ActionsBlock.java
@@ -1,0 +1,22 @@
+package com.github.seratch.jslack.api.model.block;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import com.github.seratch.jslack.api.model.block.element.BlockElement;
+
+import lombok.Builder;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+/**
+ * https://api.slack.com/reference/messaging/blocks#actions
+ */
+@Data
+@Builder
+public class ActionsBlock implements LayoutBlock {
+    private final String type = "actions";
+    @Builder.Default
+    private List<BlockElement> elements = new ArrayList<>();
+    private String blockId;
+}
--- a/src/main/java/com/github/seratch/jslack/api/model/block/LayoutBlock.java
+++ b/src/main/java/com/github/seratch/jslack/api/model/block/LayoutBlock.java
@@ -1,0 +1,18 @@
+package com.github.seratch.jslack.api.model.block;
+
+/**
+ * Block Kit is a new UI framework that offers you more control and flexibility
+ * when building messages for Slack. Comprised of "blocks," stackable bits of
+ * message UI, you can customize the order and appearance of information
+ * delivered by your app in Slack.
+ * 
+ * @see <a href="https://api.slack.com/block-kit">Block Kit Guide</a>
+ * @see <a href="https://api.slack.com/reference/messaging/blocks">Block Kit Reference</a>
+ */
+public interface LayoutBlock {
+    /**
+     * Determines the type of layout block, e.g. section, divider, context, actions
+     * and image.
+     */
+    public String getType();
+}
--- a/src/main/java/com/github/seratch/jslack/api/methods/request/chat/ChatPostMessageRequest.java
+++ b/src/main/java/com/github/seratch/jslack/api/methods/request/chat/ChatPostMessageRequest.java
@@ -2,6 +2,8 @@ package com.github.seratch.jslack.api.methods.request.chat;
 
 import com.github.seratch.jslack.api.methods.SlackApiRequest;
 import com.github.seratch.jslack.api.model.Attachment;
+import com.github.seratch.jslack.api.model.block.LayoutBlock;
+
 import lombok.*;
 
 import java.util.List;
@@ -50,6 +52,11 @@ public class ChatPostMessageRequest implements SlackApiRequest {
      * Find and link channel names and usernames.
      */
     private boolean linkNames;
+
+    /**
+     * A JSON-based array of structured blocks, presented as a URL-encoded string.
+     */
+    private List<LayoutBlock> blocks;
 
     /**
      * A JSON-based array of structured attachments, presented as a URL-encoded string.
--- a/src/main/java/com/github/seratch/jslack/api/model/block/ContextBlock.java
+++ b/src/main/java/com/github/seratch/jslack/api/model/block/ContextBlock.java
@@ -1,0 +1,20 @@
+package com.github.seratch.jslack.api.model.block;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import lombok.Builder;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+/**
+ * https://api.slack.com/reference/messaging/blocks#context
+ */
+@Data
+@Builder
+public class ContextBlock implements LayoutBlock {
+    private final String type = "context";
+    @Builder.Default
+    private List<ContextBlockElement> elements = new ArrayList<>();
+    private String blockId;
+}
--- a/src/main/java/com/github/seratch/jslack/api/model/block/ContextBlockElement.java
+++ b/src/main/java/com/github/seratch/jslack/api/model/block/ContextBlockElement.java
@@ -1,0 +1,12 @@
+package com.github.seratch.jslack.api.model.block;
+
+import com.github.seratch.jslack.api.model.block.composition.TextObject;
+import com.github.seratch.jslack.api.model.block.element.ImageElement;
+
+/**
+ * Specific interface to make context layout blocks' {@link ContextBlock
+ * elements} type-safe, because ContextBlock can only contain
+ * {@link ImageElement} and {@link TextObject} elements
+ */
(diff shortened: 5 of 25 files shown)


src/main/java/com/github/seratch/jslack/Slack.java
```diff
@@ -63,6 +64,21 @@ public WebhookResponse send(String url, Payload payload) throws IOException {
                 .build();
     }
 
+    /**
+     * Send a data to Incoming Webhook endpoint.
+     */
+    public WebhookResponse send(String url, BlockMessagePayload payload) throws IOException {
```
seratch: I don't see any necessity to have another payload class here. We can add the `blocks` property to the existing one. Also, updating the class/field comments would be preferable (if it's necessary).

src/main/java/com/github/seratch/jslack/api/model/block/ActionsBlock.java
```diff
@@ -0,0 +1,23 @@
...
+import lombok.Builder;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+/**
+ * https://api.slack.com/reference/messaging/blocks#actions
+ */
+@Data
+@EqualsAndHashCode(callSuper = true)
+@Builder
```
seratch: Regarding any classes under the `api.model` package, I intentionally avoid using `@Builder` annotation because just having it results in skipping generating default constructors for the class. 

Most JSON parsing libraries (we currently use gson internally) expect non-arg constructors. Therefore, if we simply have `@Builder`, a problem similar to the following JPA related one may happen. https://stackoverflow.com/questions/40516058/using-lomboks-data-and-builder-on-entity

I would like to explicitly have NoArgsConstructor/AllArgsConstructor as a workaround.

seratch: Also, I prefer @Builder(toBuilder = true). https://projectlombok.org/features/Builder

src/main/java/com/github/seratch/jslack/api/model/block/ActionsBlock.java
```diff
@@ -0,0 +1,23 @@
...
+public class ActionsBlock extends LayoutBlock {
+	private final String type = "actions";
```
seratch: nit: in this project, we use soft tabs with 4 spaces.

src/main/java/com/github/seratch/jslack/api/model/block/LayoutBlock.java
```diff
@@ -0,0 +1,9 @@
+package com.github.seratch.jslack.api.model.block;
+
+/**
+ * Base building block for Block Kit from Slack.
+ * 
+ * https://api.slack.com/reference/messaging/blocks
+ */
+public abstract class LayoutBlock {
```
seratch: Any reasons to choose abstract class over interface here?

dashorst: I've toyed with both variants. Either would be possible. Using a base class would allow to move shared properties from subclasses to LayoutBlock (and BlockElement, and ...). However I did not decide yet, as I haven't worked with Lombok before and:
- did not know if the properties would get the proper builder support.
- the documentation on api.slack.com mentions all properties for each class, repeating the shared ones, so I figured that mirroring the documentation might be preferable over a minimalistic OO model

I decided for abstract base class because these blocks have an is-a relationship, not an implements. That said, I can be persuaded to change it to interfaces: I'm not married to them being in a class relationship.

src/main/java/com/github/seratch/jslack/api/model/block/ActionsBlock.java
dashorst: Do you mean indentation with spaces at 4? (I love it, but need confirmation to make Eclipse conform to your coding style)

seratch: Yes, I do. It's also okay for me to apply my code formatter (I use IntelliJ IDEA's default one) when merging your changes.

src/main/java/com/github/seratch/jslack/api/model/block/LayoutBlock.java
seratch: Thank you for taking time for the clear explanation. I personally prefer interfaces here. I appreciate it if you could respect my preference.

As far as I know, we can say the possible common attribute is only `type`. Therefore, I think the interface may be like this.
```java
public interface LayoutBlock {

    /**
     * The type of block.
     **/
    String getType();
}
```

Ideally, I would like to use the explanation in `slack-api-specs` project but [it seems that the project has not supported Block Kit yet](https://github.com/slackapi/slack-api-specs/issues/18
). So, we should use the explanation in [the documentation page](https://api.slack.com/block-kit).

src/main/java/com/github/seratch/jslack/api/model/block/ActionsBlock.java
dashorst: Looking at the classes my guess is typically 4 actual spaces. 👍

src/main/java/com/github/seratch/jslack/api/model/block/ActionsBlock.java
dashorst: I've added the `@NoArgsConstructor @AllArgsConstructor` to the classes. 

I have not found a single use of `toBuilder=true` in the code base _yet_. I would like to add this in a new issue (e.g. "Add Lombok's toBuilder for easier templating of model objects").

I looked at the `api.model` package and found many `@Builder` annotations. I don't think I can find a particular pattern in which classes should/shouldn't have `@Builder` annotations.

- `@Data @Builder public class Dialog`
- `@Data @Builder public class DialogOption`
- `@Data public class DialogResponseMetadata`
- `@Data @Builder public class DialogSelectElement`
- `@Data @Builder public class Action`
- `@Data @Builder public class Attachment`
- `@Data public class BillableInfo`

I'm not criticising, but trying to understand when to apply `@Builder` and when not. Is the distinction that for objects that are returned from the Slack API you don't apply `@Builder`, but for objects that are only sent so Slack in a payload you do?

src/main/java/com/github/seratch/jslack/Slack.java
dashorst: I've removed the `BlockMessagePayload` class and corresponding method

src/main/java/com/github/seratch/jslack/api/model/block/ActionsBlock.java
dashorst: I've updated the modified classes to use spaces instead of tabs.

src/main/java/com/github/seratch/jslack/api/model/block/LayoutBlock.java
dashorst: Fixed. I've also updated the javadoc to point to the Block Kit Guide and the Block Kit Reference

src/main/java/com/github/seratch/jslack/api/model/block/ActionsBlock.java
seratch: Okay, you're right. The above comment is my intention for the future. never mind.

seratch:
TODO: I don't have time to work on writing tests for this but we need unit tests before merging this.

dashorst:
I've created a separate branch in my fork, so I'm closing this PR and submit a new one, including a unit test.

That makes it also easier to review all the changes with a fresh eye.

Thanks for helping me!

PR #101, opened 2019-03-10
Removes specific workspace assumptions from tests

### LINKED ISSUE #100: Some tests fail because of non-specific user selection
Some tests failed because the slackbot user was selected instead of an
actual user or bot. Apparently you can't invite the slackbot to a channel,
or ask about its bot-info.

dashorst:
Some tests failed because the slackbot user was selected instead of an
actual user or bot. These tests have been fixed by looking up an actual
user or bot and using that for the test case.

Fixes #100


seratch:
Oops, sorry about this. LGTM.

PR #103, opened 2019-03-12
Consistently use List over arrays, avoid nullable boolean values

seratch:
This pull request brings some additional changes to pull request #102 

@dashorst Please allow me to modify the code to make the code consistent with existing code and meet my personal preference. I've already merged the PR and even released a new version but I would appreciate it if you could understand my decision.

### DIFF
--- a/src/main/java/com/github/seratch/jslack/api/model/block/SectionBlock.java
+++ b/src/main/java/com/github/seratch/jslack/api/model/block/SectionBlock.java
@@ -7,7 +7,6 @@ import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
 
-import java.util.ArrayList;
 import java.util.List;
 
 /**
@@ -21,7 +20,6 @@ public class SectionBlock implements LayoutBlock {
     private final String type = "section";
     private TextObject text;
     private String blockId;
-    @Builder.Default
-    private List<TextObject> fields = new ArrayList<>();
+    private List<TextObject> fields;
     private BlockElement accessory;
 }
--- a/src/main/java/com/github/seratch/jslack/api/model/block/SectionBlock.java
+++ b/src/main/java/com/github/seratch/jslack/api/model/block/SectionBlock.java
@@ -7,6 +7,7 @@ import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
 
+import java.util.ArrayList;
 import java.util.List;
 
 /**
@@ -20,6 +21,7 @@ public class SectionBlock implements LayoutBlock {
     private final String type = "section";
     private TextObject text;
     private String blockId;
-    private List<TextObject> fields;
+    @Builder.Default
+    private List<TextObject> fields = new ArrayList<>();
     private BlockElement accessory;
 }
--- a/src/main/java/com/github/seratch/jslack/api/model/block/composition/MarkdownTextObject.java
+++ b/src/main/java/com/github/seratch/jslack/api/model/block/composition/MarkdownTextObject.java
@@ -13,5 +13,14 @@ import lombok.*;
 public class MarkdownTextObject extends TextObject {
     private final String type = "mrkdwn";
     private String text;
-    private boolean verbatim;
+
+    /**
+     * The documentation of the Slack API states that the verbatim field is optional.
+     * The API examples always render the emoji field (as true, but that is its default value) -- so that is not helpful.
+     * I picked the Boolean because basically you have 3 possible states:
+     * - true
+     * - false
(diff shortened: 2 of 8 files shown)


src/main/java/com/github/seratch/jslack/api/model/block/composition/OptionGroupObject.java
```diff
@@ -14,5 +17,6 @@
 @AllArgsConstructor
 public class OptionGroupObject {
     private PlainTextObject label;
-    private OptionObject[] options;
+    @Builder.Default
+    private List<OptionObject> options = new ArrayList<>();
```
dashorst: Good catch! I thought I had changed all arrays to lists.

src/main/java/com/github/seratch/jslack/api/model/block/SectionBlock.java
```diff
@@ -20,6 +21,7 @@
     private final String type = "section";
     private TextObject text;
     private String blockId;
-    private List<TextObject> fields;
+    @Builder.Default
+    private List<TextObject> fields = new ArrayList<>();
```
dashorst: The reason I didn't initialize the `List<>` fields is because Slack (sometimes) gives errors when it encounters an empty list in the generated JSON. The proper solution if/when you don't want un-initialized fields is to have a Gson type adapter that skips empty lists (e.g. something like https://stackoverflow.com/questions/11942118/how-do-you-get-gson-to-omit-null-or-empty-objects-and-empty-arrays-and-lists)

src/main/java/com/github/seratch/jslack/api/model/block/composition/MarkdownTextObject.java
```diff
@@ -13,5 +13,5 @@
 public class MarkdownTextObject extends TextObject {
     private final String type = "mrkdwn";
     private String text;
-    private Boolean verbatim;
+    private boolean verbatim;
```
dashorst: The documentation of the Slack API states that the `verbatim` field is optional (similarly the `emoji` field from `PlainTextObject` is also optional). The API examples always render the `emoji` field (as `true`, but that is its default value)--so that is not helpful. I picked the `Boolean` because basically you have 3 possible states:
- true
- false
- not present (and therefore not rendered in the resulting JSON sent to the Slack API)

src/main/java/com/github/seratch/jslack/api/model/block/element/OverflowMenuElement.java
```diff
@@ -15,6 +18,7 @@
 public class OverflowMenuElement extends BlockElement {
     private final String type = "overflow";
     private String actionId;
-    private OptionObject[] options;
+    @Builder.Default
+    private List<OptionObject> options = new ArrayList<>();
```
dashorst: Again, good catch.

src/main/java/com/github/seratch/jslack/api/model/block/element/StaticSelectElement.java
```diff
@@ -18,8 +21,10 @@
     private final String type = "static_select";
     private PlainTextObject placeholder;
     private String actionId;
-    private OptionObject[] options;
-    private OptionGroupObject[] optionGroups;
+    @Builder.Default
+    private List<OptionObject> options = new ArrayList<>();
+    @Builder.Default
+    private List<OptionGroupObject> optionGroups = new ArrayList<>();
```
dashorst: Again, good catch

src/main/java/com/github/seratch/jslack/api/model/block/composition/MarkdownTextObject.java
seratch: Right, fair enough. I will revert this change.

seratch: Will update the code with a short comment

src/main/java/com/github/seratch/jslack/api/model/block/SectionBlock.java
seratch: I understand it. I will revert the change with a comment mentioning the reason. Thanks!

dashorst:
The arrays to `List`/`ArrayList<>` was an omission on my part. I thought I had moved them all to `List`. Good catch.

@seratch I have no objections to your modifications. It is your project after all and we make different design decisions. I focused on replicating the Slack API truthfully, but that might not make a good Java API. I've added my comments inline with my design choices--this way you know why I picked `Boolean` instead of `boolean`, or did not initialise `List` fields. You're welcome to make different tradeoffs and build a more friendly Java API (without the need for having to check for `null`s).

PR #125, opened 2019-04-16
Allow the original message to be deleted or replaced

cjcdoomed:
https://stackoverflow.com/questions/47496747/how-to-remove-ephemeral-messages


seratch:
Thank you for your suggestion. I'd like to understand your intention more.

I may be wrong but, in my understanding, the field can be placed as a part of a message object in payloads. https://api.slack.com/docs/interactive-message-field-guide

Is it also possible to have it as a part of the BlockKit elements as you added? Have you already confirmed adding this property in ButtonElement works as you expect? I was not able to find any relevant documents on it.

Issue #126, opened 2019-04-17
InputStream from response


namedgraph:
Hi, how can I get an `InputStream` from `ConversationsHistoryResponse` (or any `Response` really)? I want to read raw JSON data.

namedgraph:
Example from Google's api-client-library:
https://developers.google.com/api-client-library/java/google-api-java-client/reference/1.20.0/com/google/api/client/googleapis/services/AbstractGoogleClientRequest#executeAsInputStream()

seratch:
You can access raw okhttp responses if you write some similar code to this internal method. https://github.com/seratch/jslack/blob/1.4.3/src/main/java/com/github/seratch/jslack/api/methods/impl/MethodsClientImpl.java#L1437

seratch:
okhttp’s Response class offers a method to extract request body data as input stream.

namedgraph:
I'm looking into it. Some remarks:

I think if `ChannelsHistoryRequest` and similar would be based on a map and extend `FormBody.Builder`, there would be no need to do the mappings between the two in `MethodsClientImpl` and the code would be much more generic.

I think Google client's approach

    Request req = service.method();
    Response resp = req.execute();

is more flexible than

    Request req = Request.builder()...build();
    Response resp = service.method().execute(req);

because it allows the `Request` to provide different forms of results without affecting the service interface.

namedgraph:
I wrote such a method

    public static InputStream buildResponse(Response response) throws IOException, SlackApiException {
        if (response.code() == 200) {
            return response.body().byteStream();
        } else {
            String body = response.body().string();
            throw new SlackApiException(response, body);
        }
    }

but looks like it's too much work converting `ConversationsHistoryRequest` to `ConversationsHistoryResponse` so that I can feed it into the method above.

I think a more generic design would allow more flexibility here.

seratch:
@namedgraph What do you think about #127 ? I'm going to merge the PR soon.

seratch:
Please allow me to close this issue by #127.

PR #127, opened 2019-04-17
Enable to access raw response in MethodsClient

seratch:
This pull request brings the changes which enable library users to access raw okhttp3 response objects. ref #126 cc @namedgraph

### DIFF
--- a/src/main/java/com/github/seratch/jslack/api/methods/MethodsClient.java
+++ b/src/main/java/com/github/seratch/jslack/api/methods/MethodsClient.java
@@ -98,6 +98,9 @@ import com.github.seratch.jslack.api.methods.response.usergroups.users.Usergroup
 import com.github.seratch.jslack.api.methods.response.users.*;
 import com.github.seratch.jslack.api.methods.response.users.profile.UsersProfileGetResponse;
 import com.github.seratch.jslack.api.methods.response.users.profile.UsersProfileSetResponse;
+import okhttp3.FormBody;
+import okhttp3.MultipartBody;
+import okhttp3.Response;
 
 import java.io.IOException;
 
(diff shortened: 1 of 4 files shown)


namedgraph:
It addresses the immediate need.

But I would suggest a broader refactoring. `MethodsClientImpl` is a bad design pattern - it's basically a mapping collection which suggests the logic is misplaced. It should be a generic class that does not know about parameters. Try to move parameters into the `Request` classes and make them extend `Builder`:

    public class ConversationsHistoryRequest extends FormBody.Builder implements SlackApiRequest {

        private final Form.Builder builder;

        private ConversationsHistoryRequest() {
            builder = new Builder();
        }

        public ConversationsHistoryRequest new() {
            return new ConversationsHistoryRequest();
        }
        
        public ConversationsHistoryRequest channel(String channel) {
            if (token != null) builder.add("channel", channel);
            else builder.remove("channel");
            return this;
        }

        public ConversationsHistoryRequest token(String token) {
            if (token != null) builder.add("token", token);
            else builder.remove("token");
            return this;
        }

        ...

    }

Then `MethodsClientImpl` will need only few generic methods:

    public class MethodsClientImpl implements MethodsClient {

            protected <T> T doPostForm( ...

            protected <T> T doPostFormWithToken(...

            protected <T> T doPostMultipart(...

    }

And the invocation looks simply like this:

    ConversationsHistoryRequest req = ConversationsHistoryRequest.new().
        channel("C1234567890").
        token("xxxx-xxxxxxxxx-xxxx");
    MethodsClient client = new MethodsClientImpl();
    client.doPostForm(req, ...);

seratch:
I understand your point but I'm not willing to inherit `FormBody.Builder`. To me, having both setters and `add(key, value)` method as public APIs is not clean.

namedgraph:
That is the nature of a builder.

My point is that `MethodsClientImpl ` is a bottleneck in your codebase which prevents it from being more generic/functional.

It's your project, do as you wish :)

seratch:
Thank you for your feedback and also understanding. I’m going to ship a new version including this change this weekend.

PR #137, opened 2019-04-25
Add chat scheduled messages methods

seratch:
This pull request adds supports for scheduled messages support under `chat` namespace.

Also, I've added some methods under `apps.permissions`. They're workspace apps related ones. So, actually, we don't need to support them anymore, though.


src/main/java/com/github/seratch/jslack/api/methods/RequestFormBuilder.java
```diff
@@ -227,19 +269,63 @@ private RequestFormBuilder() {
...
+        return form;
+    }
+
     public static FormBody.Builder toForm(ChatPostEphemeralRequest req) {
         FormBody.Builder form = new FormBody.Builder();
         setIfNotNull("channel", req.getChannel(), form);
         setIfNotNull("text", req.getText(), form);
         setIfNotNull("user", req.getUser(), form);
         setIfNotNull("as_user", req.isAsUser(), form);
+        if (req.getBlocks() != null) {
```
seratch: It was an existing bug. I overlooked this for some reason.

PR #139, opened 2019-04-29
Don't kill thread when event handling fails

sean-heller:
Currently , if the EventsDispatcher fails deserialize an event payload, or if an exception occurs during the implementing handler, the thread polling the queue will die and not continue polling. This means you'll miss all events after the first failure. Instead, it should log an error but continue handling events.


seratch:
@sean-heller Thank you for your contribution. This change is included in the 1.5.3 release I published yesterday. Please try it out when you have a chance.

PR #152, opened 2019-05-11
Add missing params to MessageEvent

sean-heller:
See [docs for bot_message subtype](https://api.slack.com/events/message/bot_message)

Not sure if this is the right place, or if there are subtypes defined elsewhere


seratch:
Thanks, LGTM. It seems that `icons` is also missing but I will work on it after merging this PR. Thank you for your contribution!

PR #155, opened 2019-05-13
Split the library into several modules (aiming Android support)

seratch:
Since jSlack 1.6, I'm going to split this library into the following submodules. The main purpose is to enable using this library for Android app development.

* jslack (all-in-one package) - Java8+
* jslack-api-model (API data models) - Java7+
* jslack-common (commonly used implementation) - Java7+
* jslack-api-client (Android-ready API client) - Java7+
* jslack-app-backend (Features for Slack App development) - Java8+
* jslack-shortcut (jSlack's original handy APIs) - Java8+
* jslack-test-utils (only used for this library's tests) - Java7+

## Breaking changes

This pull request brings the following breaking change to Shortcut APIs. These are inevitable.

* `Slack.shortcut()` to `new SlackShortcut(Slack.getInstance()).getInstance()`
* `Slack.shortcut(token)` to `new SlackShortcut(Slack.getInstance()).getInstance(ApiToken.of(token))`


seratch:
I overlooked that okhttp 3 recently dropped Java7 support. We have two options below:

* 1) Give up supporting Java7 (= older Android version), make jslack builds simpler
* 2) Downgrade okhttp to 3.12 series + support Java7

I'm thinking to go with plan 1) so far.

seratch:
I decided to go with plan 1. The module structure is now much simpler. Plus, we don't bring any breaking changes this time 🎉 If a user wants to load all the existing modules, just specify artifactId `jslack` as they have been up to now. If a user would like to load Slack API client, use artifactId `jslack-api-client` (which depends on `jslack-api-model`). From the viewpoint of Android's 64K method limit, this improvement may be beneficial for some people.

* `jslack` (all-in-one package - backward compatible)
* `jslack-api-model` (API data models)
* `jslack-api-client` (Android-ready API client) 
* `jslack-app-backend` (Collection of the features for Slack App development)

PR #174, opened 2019-06-28
fix(channel-cache): follow pagination cursor

FrankSpitulski:
this lets the channel cache be filled for large workspaces with over 1000 channels


seratch:
Sorry for my belated response. Looks good to me. Thank you for your contribution!

Issue #175, opened 2019-07-02
Missing Guest Channels Properties in User Profile


linlei99:
Hi,

We have some invited guest users in our Slack. Two properties below are missing in the User Profile class which is causing Gson fails to convert the response Json string to Slack User objects.

"guest_channels" and "guest_invited_by" 

Could you please add them in User.java in the master branch so that we can keep our project going? 

Thank you so much.

Lei

seratch:
Thanks for your pull request on this #176 
I will release a new version including the change very soon.

seratch:
version 1.7.3 is out. Thank you for your feedback and contributions!

linlei99:
Thank you so much!

PR #176, opened 2019-07-04
-- added missing properties.

linlei99:
The fix is for #175. Please review it. Thank you.

### DIFF
--- a/jslack-api-model/src/main/java/com/github/seratch/jslack/api/model/User.java
+++ b/jslack-api-model/src/main/java/com/github/seratch/jslack/api/model/User.java
@@ -49,6 +49,8 @@ public class User {
     @Data
     public static class Profile {
 
+        private String guestChannels;
+        private String guestInvitedBy;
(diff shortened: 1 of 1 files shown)


jslack-api-model/src/main/java/com/github/seratch/jslack/api/model/User.java
```diff
@@ -49,6 +49,8 @@
     @Data
     public static class Profile {
 
+        private String guestChannels;
```
seratch: I haven’t checked the actual response yet but is it a string value? Not an array?

seratch: Just made sure it and I found it's a string value having an array inside.
```
        "guest_channels": "[\"C88888888\",\"C88888888\",\"C88888888\"]",
        "guest_invited_by": "U88888888",
```

seratch:
Thanks!

PR #188, opened 2019-07-22
Fix #187: Attachment blocks result in invalid_attachments response

### LINKED ISSUE #187: Attachment blocks result in invalid_attachments response
Attachment blocks (#157) don't work as the JSONified Attachment includes boolean `"indent":false` which causes Slack to give an invalid_attachments response. Thought I had got all of them, but think this got added in the latest release :(

Working on a solution similar to #184

domdomegg:
Sorry, I missed one that was added recently.


seratch:
Oops, I also missed it. Thanks. Will release a new version soon.

seratch:
jSlack version 1.7.6 is available on the Sonatype repository now. It'll be synched to Maven Central soon.

PR #192, opened 2019-08-01
Improve SlackEventsApiServlet.java to verify X-Slack-Signature header (fixes #163 )

### LINKED ISSUE #163: Question: How do you verify the payload coming from EventsAPI?
Congratulations for the API it's very feature complete. I got one doubt on origin verification. You know you can verify requests on slack according to their practices: https://api.slack.com/docs/verifying-requests-from-slack

Maybe I just missed some part of the code but does the code for EventsAPI get the request headers and do that validation? Before calling the handlers?

Thanks!

shubhvashisht:
fixed the test case.

### DIFF
--- a/jslack-app-backend/src/main/java/com/github/seratch/jslack/app_backend/events/servlet/SlackEventsApiServlet.java
+++ b/jslack-app-backend/src/main/java/com/github/seratch/jslack/app_backend/events/servlet/SlackEventsApiServlet.java
@@ -13,7 +13,10 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 import java.nio.charset.Charset;
+import java.util.Enumeration;
+import java.util.HashMap;
 import java.util.Locale;
+import java.util.Map;
 import java.util.stream.Collectors;
 
 @Slf4j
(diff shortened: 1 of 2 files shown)


seratch:
Thank you for making a pull request for this issue. Your change seems to be:
* This one doesn't verify `X-Slack-Signature` header yet (hint: SlackSignatureVerifier)
* I would like to avoid having unrelated changes like creating a Map object containing all request headers
* Some related tests are failing - check the CI build results
* Please apply code formatter (ideally, IntelliJ IDEA's default one) after pushing your code

seratch:
Thank you for making a pull request for this issue. Let me suggest:
* This one doesn't verify `X-Slack-Signature` header yet (hint: SlackSignatureVerifier)
* I would like to avoid having unrelated changes like creating a Map object containing all request headers
* Some related tests are failing - check the CI build results
* Please apply code formatter (ideally, IntelliJ IDEA's default one) before pushing your code

seratch:
ref #163

shubhvashisht:
> This one doesn't verify X-Slack-Signature header yet (hint: SlackSignatureVerifier)

Yeah, for now it is not doing slack signature verify. I was thinking that probably client can implement it and then call this function. But as I think more, I think an optional parameter will be helpful for anyone who wants to use this feature with signed secret. I will work on it.

> I would like to avoid having unrelated changes like creating a Map object containing all request headers

Yes, i realized that and have removed the same.

> Some related tests are failing - check the CI build results
Already fixed.

seratch:
>I think an optional parameter will be helpful for anyone who wants to use this feature with signed secret.

Good point. 1) having a constructor which requires signing secret or 2) defining a getter `String getSlackSignedSecret()` may be reasonable. I prefer going with 1)

shubhvashisht:
closing the pull request as added new features.

PR #193, opened 2019-08-02
Improve SlackEventsApiServlet.java to verify X-Slack-Signature header (fixes #163 )

shubhvashisht:

PR #194, opened 2019-08-03
Add X-Slack-Signature verification in SlackEventsApiServlet

seratch:
This pull request simplifies #193 and added some improvements for the signature verification utilities. 

@shubhvashisht I had to change your code a little bit but your suggestion was really helpful! Thank you for your great contribution to this library. 👍

### DIFF
--- a/jslack-app-backend/src/main/java/com/github/seratch/jslack/app_backend/SlackSignature.java
+++ b/jslack-app-backend/src/main/java/com/github/seratch/jslack/app_backend/SlackSignature.java
@@ -35,6 +35,26 @@ public class SlackSignature {
         public static final String DEFAULT_ENV_NAME = "SLACK_SIGNING_SECRET";
     }
 
+    public static class TimestampVerifier {
+
+        public static final int TIMESTAMP_EXPIRATION_TIME_IN_MILLIS = 60 * 5 * 1000; // 5 minutes
+
+        private TimestampVerifier() {
+        }
+
+        public static boolean isValidTimestamp(String timestampValue, long nowInMillis) {
+            try {
+                long requestMillis = Long.parseLong(timestampValue) * 1000;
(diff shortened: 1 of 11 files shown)


shubhvashisht:
@seratch thanks for merging the pull request.

Few things to consider here for the code refactor:
- A single jslack library can be used by multiple apps on the same server. This implies that each app can have its own secret key for signed secret validation. Therefore, instead of system environment variable if we have it as a incoming parameter then it can serve multiple applications on the same server.

- The property of enabling the signed secret should instead have a getter/setter wherein the client can specify whether validation is required. This may help the enterprise apps wherein the signed secret validation is done at the API gateway layer instead of the application layer. By this implementation, the apps can determine when to do signed secret validation.

Thoughts?

seratch:
@shubhvashisht 
 1) Good point. This is obviously the point we should improve current implementation. I will work on it.

2) I don't have any plans to work on features which may be equivalent to the things here. The API servlet implementation jSlack offers is an abstract class and library users can freely override the getter method. Even if a library user would like to configure the flag, it doesn't necessarily mean having a setter is required.

3) My intention of `SlackEventsApiServlet ` is to provide a simple base class. If the base class doesn't meet your demands, please implement your own one.

In summary, I will improve `SlackEventsApiServlet` to consider multiple workspace apps while I won't work on the rest. It would be appreciated if you could understand my stance.

seratch:
>Good point. This is obviously the point we should improve current implementation. I will work on it.

Ah, no. I was somewhat confused. If a simple servlet app doesn't need to switch multiple Slack app settings dynamically, the current implementation should work fine. Library users can configure signing secret used in servlet apps by overriding the method and can have multiple secrets for multiple servlets in a single server.

shubhvashisht:
Got it.

PR #203, opened 2019-08-11
Added a new payload and handler for app_home_open

d-subrahmanyam:
Added a new payload and handler for app_home_open


d-subrahmanyam:
When I was trying to create and use this class locally, I was getting a ClassCastException.

seratch:
@d-subrahmanyam Good catch. Thank you for making a PR! I'm going to release a new version including this fix today.

> When I was trying to create and use this class locally, I was getting a ClassCastException.

If you still have the error message / stacktrace, could you share it with us just for reference?

seratch:
I just shipped a new version 1.8.1. Thanks again 👍

d-subrahmanyam:
Sure. Thanks.

PR #204, opened 2019-08-13
modified the ChatPostMessage to set blocks as string

d-subrahmanyam:
modified the ChatPostMessage to set blocks as string. This is to enable setting the blocks text as a string read from a json template. This would be helpful while building bots with many block messages.


seratch:
Thanks for your contribution again! I've merged your idea with a few changes.

* Renamed blocksText to blocksAsString (text could be a bit confusing with `text`)
* Add the same supports for chat.postEphemeral, chat.update, chat.scheduleMessage

d-subrahmanyam:
Sure. Thanks a lot. I was thinking, how about we use something like Netty/Akka in place of the existing SlackApiServlet. I will work on it over the weekend and raise a PR.

d-subrahmanyam:
Is there way, I could reach out to you direct? like Slack/Email?

seratch:
>how about we use something like Netty/Akka in place of the existing SlackApiServlet. I will work on it over the weekend and raise a PR.

I don't have plans to offer an Akka module in this project. Could you consider starting it as your own project?

To outreach more developers, it may be great to support Spring Boot. Even in that case, I'll consider starting a new project which depends on this project.

d-subrahmanyam:
I think Spring Boot or Vertx would also be good candidate

PR #213, opened 2019-09-03
subtype property for the app_mention event

d-subrahmanyam:
- Added a subtype property to the app_mention event
- Updated the relevant testcase and json


seratch:
Thanks, LGTM

PR #251, opened 2019-11-04
Add missing types of rich sections

k-sever:
See https://api.slack.com/changelog/2019-09-what-they-see-is-what-you-get-and-more-and-less
Only `text` and `emoji` are supported atm, added the rest of types.


seratch:
Thanks!

PR #257, opened 2019-11-09
Fix postAsBot to actually post as bot, not as user

### LINKED ISSUE #256: Shortcut.postAsBot actually posts as user
Hi, 
I've noticed that trying to post message formatted with `List<LayoutBlock>` as bot, what actually happens is that message is posted as user.

Issue seem to be on this line:
https://github.com/seratch/jslack/blob/f54633c369697a1898ea526549651e02b74770db/jslack-api-client/src/main/java/com/github/seratch/jslack/shortcut/impl/ShortcutImpl.java#L154
last argument is `true` (boolean asUser) but should be `false`

antonio-tomac:
Fixes #256


seratch:
Ah, silly mistake. Thanks!

PR #258, opened 2019-11-11
Ensure team and enterprise ids are set on requests

jmccance:
Seems that Requests are responsible for updating their contexts, but only some of them were updating the enterprise and team ids. This is an issue since MultiTeamsAuthorization is expecting these fields in order to find the right bot user.

This goes through all the request types that seem to have access to team id and ensures that it (and it's undocumented buddy enterprise id, if available) are always added to the request context.

### DIFF
--- a/jslack-lightning/src/main/java/com/github/seratch/jslack/lightning/request/builtin/EventRequest.java
+++ b/jslack-lightning/src/main/java/com/github/seratch/jslack/lightning/request/builtin/EventRequest.java
@@ -23,6 +23,7 @@ public class EventRequest extends Request<DefaultContext> {
         this.headers = headers;
         JsonObject payload = GsonFactory.createSnakeCase().fromJson(requestBody, JsonElement.class).getAsJsonObject();
         this.eventType = payload.get("event").getAsJsonObject().get("type").getAsString();
+        this.getContext().setTeamId(payload.get("team_id").getAsString());
     }
 
     private DefaultContext context = new DefaultContext();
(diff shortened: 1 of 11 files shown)


jslack-lightning/src/main/java/com/github/seratch/jslack/lightning/request/builtin/EventRequest.java
```diff
@@ -23,6 +23,7 @@ public EventRequest(
         this.headers = headers;
         JsonObject payload = GsonFactory.createSnakeCase().fromJson(requestBody, JsonElement.class).getAsJsonObject();
         this.eventType = payload.get("event").getAsJsonObject().get("type").getAsString();
+        this.getContext().setTeamId(payload.get("team_id").getAsString());
```
seratch: We can set enterprise_id too. I'll add it after merging this PR.

seratch:
@jmccance Thanks again for your contributions! 🙏  I just released version 3.0.4 onto the Sonatype OSS repository. The jars will be synced to Maven Central very soon.

jslack-lightning/src/main/java/com/github/seratch/jslack/lightning/request/builtin/EventRequest.java
jmccance: Good call.

jmccance:
Excellent. Thanks for your quick reviews and releases!

PR #280, opened 2019-11-20
Remove Providers

jeffbrown:


seratch:
Thanks, this is really cool! I'm still new to Micronaut. I learned from you 👍

PR #294, opened 2019-12-05
Fix #293 SlackApiException and other exceptions are not standard

### LINKED ISSUE #293: SlackApiException and other exceptions are not standard
SlackApiException does not set Throwable.detailMessage, so when you catch the exception and try to log it, the message is "null".

Since you cannot set the message by calling `Exception(String message)` would it be possible to overwrite Throwable::getMessage() to produce a readable message when logging?

seratch:
This pull request improves the issue pointed out by #293.

@kushal256 Could you leave comments if you see something to improve more?

### DIFF
--- a/jslack-api-client/src/main/java/com/github/seratch/jslack/api/methods/SlackApiException.java
+++ b/jslack-api-client/src/main/java/com/github/seratch/jslack/api/methods/SlackApiException.java
@@ -19,8 +19,26 @@ public class SlackApiException extends Exception {
     }
 
     public SlackApiException(SlackConfig config, Response response, String responseBody) {
+        this(response, responseBody, parse(config, responseBody));
+    }
+
+    public SlackApiException(Response response, String responseBody, SlackApiErrorResponse error) {
+        super(buildErrorMessage(response, error));
         this.response = response;
         this.responseBody = responseBody;
+        this.error = error;
+    }
+
+    private static String buildErrorMessage(Response response, SlackApiErrorResponse error) {
+        String message = "status: " + response.code();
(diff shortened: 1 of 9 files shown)


jslack-api-client/src/main/java/com/github/seratch/jslack/api/methods/SlackApiException.java
```diff
@@ -19,8 +19,26 @@ public SlackApiException(Response response, String responseBody) {
...
+        super(buildErrorMessage(response, error));
         this.response = response;
         this.responseBody = responseBody;
+        this.error = error;
+    }
+
+    private static String buildErrorMessage(Response response, SlackApiErrorResponse error) {
+        String message = "status: " + response.code();
+        if (error != null) {
+            return message + ", error: " + error.getError() + ", needed: " + error.getNeeded() + ", warning: " + error.getProvided() + ", warning: " + error.getWarning();
```
kushal256: This has the text `warning` twice, i think first one should be `provided`?

Would it be easier to just make this line instead `return message + error.toString()` as SlackApiErrorResponse already has toString() implemented from the lombok `@Data` annotation?

seratch: Good catch, thanks! >warning

>Would it be easier to just make this line instead return message + error.toString() as SlackApiErrorResponse already has toString() implemented from the lombok @Data annotation?

True, it would be much easier but I don't think it's good as the string would be much longer and it would be like "SlackApiErrorResponse(....)".

I don't think it's a good way

PR #301, opened 2019-12-11
Skip team auth for URL verification requests

jmccance:
URL verification requests are teamless, so we cannot verify our connectivity to them.


seratch:
Thanks, you're right. I just merged the fix based on yours (sorry, I wanted to have your commit in history but I failed). I will ship a new version.

seratch:
I just released version 3.2.5. Thanks for pointing this issue out.

jmccance:
No prob. 🙂 Just glad to have the fix in one way or the other.

PR #304, opened 2019-12-17
Add bot_id in response for auth.test

seratch:
This pull request adds `bot_id` in response for `auth.test` API. https://api.slack.com/methods/auth.test


jslack-app-backend/src/main/java/com/github/seratch/jslack/app_backend/util/JsonPayloadExtractor.java
```diff
@@ -19,7 +19,7 @@
 
     public String extractIfExists(String requestBody) {
 
-        if (requestBody == null) {
+        if (requestBody == null || requestBody.trim().length() == 0) {
```
seratch: This one is not related to `auth.test` API updates but it seems like a potential bug.

PR #306, opened 2019-12-18
Remove conversation.members as it has not been working

seratch:
I just noticed somehow `Conversation` class has `members` field in it but it has not been working from the beginning. When API users need to know user ids that belong to a channel, calling `conversations.members` API is necessary.

I don't think this is a breaking change as it never worked correctly and nobody depends on the field.

### DIFF
--- a/jslack-api-model/src/main/java/com/github/seratch/jslack/api/model/Conversation.java
+++ b/jslack-api-model/src/main/java/com/github/seratch/jslack/api/model/Conversation.java
@@ -38,7 +38,6 @@ public class Conversation {
     private List<String> previousNames;
     @SerializedName("num_members")
     private Integer numOfMembers;
-    private List<String> members;
     private Latest latest;
     private String locale;
     @SerializedName("unread_count")


joeltoby:
Good catch.  The members field did used to work, but the API was changed.  Removing the field shouldn't be a breaking change.

FYI, the change was discussed on [Slack's eng blog here](https://slack.engineering/evolving-the-slack-api-8a329f0a5645):

> _In some cases, we changed the arguments to improve performance and scalability for apps. For example, channels.info used to include a nested list of members, but as larger and larger companies started to use Slack, response payloads grew, in some cases to as much as 290MB for a single channel. In addition, 72% of requests to channels.info specified no_members=true. To help apps scale to large teams, we now exclude members by default for conversations.info. To provide the channel membership data in a more optimal format to developers, we added a new conversations.members endpoint with pagination._

seratch:
@joeltoby 👋 Thanks for the comment! Good to know(?) that the field worked at that moment.

PR #420, opened 2020-04-06
Fix #419 Enable access to App in SlackApiLambdaHandler

### LINKED ISSUE #419: SlackApiLambdaHandler forbids access to App
### Issue Type

- [ ] Bug
- [x] Enhancement / Feature request
- [ ] Question
- [ ] Documentation

### Description

`SlackApiLambdaHandler` is very handy to quickly create a working lambda slack bot. The problem is when you extend from it you have to call it's constructor passing an `App` instance, but you don't have an easy way to modify that instance later on. I can think of a few examples where that is important:
- I would like to add a middleware on top of the default ones. When a single call to `App#use` would suffice I need to either replicate the logic of `buildDefaultMiddlewareList`, use composition instead of inheritance or use a static `App`.
- I would like to add a handler for command, for example. A call to `App#command` would be enough but I have to do some of the workarounds described previously.

I guess the idea is express and no more use cases are required. Would you consider changing this? Is there a reason why you did it this way? Even though create a working lambda would be fast with this class doing something useful would require some kind of workaround as it is.

### Requirements (place an `x` in each of the `[ ]`)

* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/java-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've searched for any related issues and avoided creating a duplicate issue.

seratch:
###  Summary

This pull request fixes #419 by adding a `protected` getter for the underlying `App` instance in `SlackApiLambdaHandler`. Although such a use case is pretty rare considering the nature of FaaS, there is no strong reason to avoid having the accessor.


bolt-aws-lambda/src/main/java/com/slack/api/bolt/aws_lambda/SlackApiLambdaHandler.java
```diff
@@ -30,6 +30,10 @@ public SlackApiLambdaHandler(App app) {
         this.requestParser = new SlackRequestParser(app.config());
     }
 
+    protected App getApp() {
```
acm19, line 33: I think a Javadoc explaining the purpose of the method would be useful. I was also thinking if it makes sense to call this method just `app` to be consistent with `App` own naming `App#config`, `App#status`, etc.

seratch: okay, i'm fine with it.

PR #421, opened 2020-04-06
Add access to `SlackApiLambdaHandler` internal `App`

acm19:
###  Summary

Adds a protected method `SlackApiLambdaHandler#app` to allow lambda
handlers implementing that class to modify the behaviour of the
underlying `App` after initialisation.

Moves the call to `App#start` to the constructor to avoid unnecessarily
call it every time a warmup event triggers the lambda. It is only needed
to run it once when the lambda starts up the first time.

Reorders the methods in `SlackApiLambdaHandler` to be next to the
calling method, enhacing readability.


bolt-aws-lambda/src/main/java/com/slack/api/bolt/aws_lambda/SlackApiLambdaHandler.java
```diff
@@ -98,4 +87,33 @@ public ApiGatewayResponse handleRequest(ApiGatewayRequest input, Context context
...
+
+    /**
+     * Allows access to the internal {@link App} used by this class interact
+     * with Slack. It is useful in cases where changes on the app are necessary
+     * after initialisation. For example, when adding a reaction after certain
+     * command calls this bot.
+     *
+     * @return the {@code App} used by this class to handle events
+     */
+    protected App app() {
```
seratch, line 116: I was thinking `app()` may be also fine here. I'm fine with `app()` instead of `getApp()`.

seratch:
Some idea will be reflected in #420

PR #441, opened 2020-05-01
Fix block element equality

eamelink:
###  Summary

Block elements should be compared structurally, but their `@EqualsAndHashCode`  annotation had `callSuper  = true`. The super class however (typically `ButtonElement`) doesn't have `equals` overridden for structural equality. So two structurally identical elements would not be considered equal by `.equals()`.

In other words,  this was failing:

    assertEquals(new ButtonElement(), new ButtonElement())

The first commit in this PR fixes that, by removing the `callSuper = flag` from the `@EqualsAndHashCode` annotation, so that these elements are considered equals when they are structurally equivalent. 

Two additional commits that don't change anything, but just clean up a  bit of code:
 - First one removes `@EqualsAndHashCode` when there's already `@Data`, which includes  that.
- Last one removes `callSuper = false` from the remaining `@EqualsAndHashCode` annotations, since it's the default value.

### DIFF
--- a/bolt/src/main/java/com/slack/api/bolt/context/builtin/ActionContext.java
+++ b/bolt/src/main/java/com/slack/api/bolt/context/builtin/ActionContext.java
@@ -14,7 +14,7 @@ import lombok.*;
 @NoArgsConstructor
 @AllArgsConstructor
 @ToString(callSuper = true)
-@EqualsAndHashCode(callSuper = false)
+@EqualsAndHashCode
 public class ActionContext extends Context implements ActionRespondUtility {
 
     private String triggerId;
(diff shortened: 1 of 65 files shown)


seratch:
@eamelink Great catch! (yes, this is my fault...) Thank you very much for your contribution 🙇

seratch:
@eamelink I've overlooked this but having `@EqualsAndHashCode(callSuper=false)` was intentional. You should see the following warnings by Lombok. Let me revert this commit: https://github.com/slackapi/java-slack-sdk/pull/441/commits/41294ce9b94ac8e373472adff22a672beda85d22

```
/java-slack-sdk/slack-api-client/src/main/java/com/slack/api/scim/response/UsersReadResponse.java:7: warning: Generating equals/hashCode implementation but without a call to superclass, even though this class does not extend java.lang.Object. If this is intentional, add '@EqualsAndHashCode(callSuper=false)' to your type.
@Data
^
```

Issue #428, opened 2020-04-10
Add optional Kotlin builders for building messages


emanguy:
### Issue Type

- [ ] Bug
- [x] Enhancement / Feature request
- [ ] Question
- [ ] Documentation

### Description

I'd like to be able to build slack messages in more of a "kotlin builder" style rather than the functional chain builders that are currently available. I'm also willing to contribute the code if need be, just want to add an issue here.

For example, here's how building a slack message currently looks in Kotlin code:
```kotlin
slack.methods(token).chatPostMessage { req -> 
  req.channel("C1234567")
  .blocks(asBlocks(
    section { section -> section.text(markdownText("*Please select a restaurant:*")) },
    divider(),
    actions { actions -> 
      actions.elements(asElements(
        button { b -> b.text(plainText(pt -> pt.emoji(true).text("Farmhouse"))).value("v1") },
        button { b -> b.text(plainText(pt -> pt.emoji(true).text("Kin Khao"))).value("v2") }
      ))
    }
  ))
);
```

If we were to use a kotlin builder style, we could do something like this instead, which fits in with the language a little better:

```kotlin
slack.methods(token).buildAndPostMessage {
  channel = "C1234567"
  blocks {
    section {
      markdownText("*Please select a restaurant:*")
    }
    divider()
    actions {
      elements {
        button {
          plainText(text = "Farmhouse", emoji = true, value = "v1")
        }
        button {
          plainText(text = "Kin Khao", emoji = true, value = "v2")
        }
      }
    }
  }
}
```

### Requirements (place an `x` in each of the `[ ]`)

* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/java-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

seratch:
@emanguy Thanks for taking the time to write in! The suggested API looks neat. One thing I'd like to know more is how the implementation looks like. Is it possible to ask you to show proof of concept code (you don't need to support many APIs)?

If it continuously requires efforts for future maintenance, I may be a bit reluctant to add the support.

emanguy:
Okay, I'll go ahead and throw some sample code into a pull request and reference here.

emanguy:
[...]

PR #469, opened 2020-05-22
Block Kit Kotlin DSL

emanguy:
###  Summary

Implementation of Kotlin DSL as mentioned in #428 


seratch:
@emanguy This is great! I will check this PR in detail tomorrow! To make the CI builds successful, can you run `./scripts/set_version.sh 1.0.12-SNAPSHOT` and push the change by the command?

emanguy:
The script didn't seem to modify the pom.xml files in the kotlin packages so I just manually changed the package version. Let's see if that makes CI happy.

seratch:
@emanguy 
>The script didn't seem to modify the pom.xml files in the kotlin packages

I'm not sure why it didn't work for you (it works for me). Let me know if you find out the reason.

Thank you very much for your great work! 

I've checked the changes and added automated tests verifying the field coverage. Those tests helped me find missing supports and inconsistent field naming. I've submitted a PR to your branch. Could you check it out when you have a chance? https://github.com/emanguy/java-slack-sdk/pull/4

After merging my changes,  I'm sure we can merge this PR. As this PR has so many commits, I'm going to squash the commits into one. If you can organize the commits to a few relevant ones, it would be greatly appreciated (you don't need to keep my commits).  But it's totally optional. I can merge this PR as a single commit on my end.

slack-api-model-kotlin-extension/src/main/kotlin/com/slack/api/model/kotlin_extension/block/composition/OptionGroupObjectBuilder.kt
```diff
@@ -27,8 +26,13 @@ class OptionGroupObjectBuilder private constructor(
...
     }
 
-    fun options(options: List<OptionObject>) {
-        this.optionContainer.underlying.addAll(options)
+    /**
+     * An array of option objects that belong to this specific group. Maximum of 100 items.
+     *
+     * @see <a href="https://api.slack.com/reference/block-kit/composition-objects#option_group">Option group object documentation</a>
+     */
+    fun options(builder: OptionObjectDsl.() -> Unit) {
```
seratch, line 34: As this method is not `option` but `options`, changing it this way is irrelevant. `OptionObjectDsl` provides only a way to construct a single `OptionObject`.

emanguy:
I just squashed sets of commits into logical changes and rebased on top of master.

slack-api-model-kotlin-extension/src/main/kotlin/com/slack/api/model/kotlin_extension/block/composition/OptionGroupObjectBuilder.kt
emanguy: I think this conversation got lost after the rebase. Is it still applicable? I was doing this to make `options()` behave the way `elements()` does on some of the block builders. For example:

```kotlin
// ...
optionGroup {
  label("Some label")
  options {
    option {
      // Option data here
    }
    option {
      // Option data here
    }
  }
}
```

Otherwise, the builder methods from OptionObjectDsl will be unavailable and the user would have to construct each option object by hand

emanguy: Edited my above comment for clarity.

seratch: Thanks for clarifying this! I was a bit confused when I left this comment. I just wrote some code with your latest revision and I understand how it works. LGTM.

seratch:
@emanguy Thanks again for your great work! This PR looks great to me. Other maintainers and watchers may have some thoughts. Just in case, I'll wait for one more day before merging this.

The next release including this PR is going to be a new minor version - 1.1.0. The ETA for the release is within a few business days (probably, by the end of this month).
https://github.com/slackapi/java-slack-sdk/milestone/2

emanguy:
Awesome! It was a lot of fun working on this, thanks for all your help and insight. Feel free to @ me if any further work needs to be done on it.

seratch:
@emanguy 
I'm sorry I have one thing I should have mentioned earlier. In short, I'd like to remove `rich_text` block support.

Currently, composing a message utilizing `rich_text` blocks is not supposed to be done by any 3rd party Slack apps. As there is no document about it apart from [this announcement](https://api.slack.com/changelog/2019-09-what-they-see-is-what-you-get-and-more-and-less), only Slack 1st party app's WYSIWYG editor can generate valid rich_text blocks. Also, the reason why this Java SDK added `rich_text` block support is the necessity to parse those blocks in `conversations.history` responses etc.

So, to avoid unnecessary confusion among developers, please allow us to remove the DSLs for `rich_text` blocks at this point. We may revisit this in the future if the Slack platform utilizes `rich_text` blocks for some other purpose, though.

I will send a pull request to your branch. Could you take a look at it? If you're too busy to check it, I can merge this and then remove the `rich_text` support afterward.

emanguy:
Sure, I'll go ahead and remove the builders. We can always revert the commit later when the API is ready.

seratch:
Thanks a lot 🙇

emanguy:
The rich text builders are removed.

emanguy:
...and now the tests are fixed. I missed a rich text reference in the client extension module 😅

seratch:
Thank you very much again for your huge contribution in this PR and the whole productive conversation in the issue thread. Now we can safely merge this and I'm really glad to have such amazing modules in this SDK!

PR #490, opened 2020-06-17
Fix #489 Event type extraction failure in slack-app-backend (no effect to Bolt)

### LINKED ISSUE #489: Event type extract does not match correctly if there are "type" properties in JSON subentries.
The current code for extracting event types seems to assume that the "type:" entry that contains the event type is the first type entry in the JSON that follows "event:". However, this is not always the case. 

E.g. in the following message `{"authed_users":["RETRACTED"],"event_id":"RETRACTED","api_app_id":"RETRACTED","team_id":"RETRACTED","event":{"client_msg_id":"b5ac4998-4526-4d0b-8bb0-7bd614c1b774","blocks":[{"elements":[{"elements":[{"text":"test","type":"text"}],"type":"rich_text_section"}],"type":"rich_text","block_id":"Ws3Y"}],"event_ts":"1592257234.003200","channel":"RETRACTED","text":"test","team":"RETRACTED","type":"message","channel_type":"channel","user":"RETRACTED","ts":"1592257234.003200"},"type":"event_callback","event_time":1592257234,"token":"RETRACTED"}` the `"type": "text"`value of the "blocks" field comes first. The event for that message is detected as "text" even though it would need to be "message" to arrive in the correct event handler.

https://github.com/slackapi/java-slack-sdk/blob/a2e96f3582189cbd7e9b0bd5a4504bdba4262ce6/slack-app-backend/src/main/java/com/slack/api/app_backend/events/EventTypeExtractorImpl.java#L23

seratch:
###  Summary

This pull request fixes #489 affecting only the users that directly use the **slack-app-backend** library. Bolt for Java does the extraction in a different way. 

Manipulating string values in a bit naive way (yes, I wrote the code!) may cause similar issues in the future. So, I've changed the implementation to simply rely on [Gson](https://github.com/google/gson) as with Bolt's approach.

### DIFF
--- a/slack-app-backend/src/test/java/test_locally/app_backend/events/EventTypeDetectorImplTest.java
+++ b/slack-app-backend/src/test/java/test_locally/app_backend/events/EventTypeDetectorImplTest.java
@@ -12,6 +12,26 @@ public class EventTypeDetectorImplTest {
     private EventTypeExtractor eventTypeExtractor = new EventTypeExtractorImpl();
 
     @Test
+    public void notFound() {
+        {
+            String eventType = eventTypeExtractor.extractEventType("{}");
+            assertThat(eventType, is(""));
+        }
+        {
+            String eventType = eventTypeExtractor.extractEventType("{\"event\": {}}");
+            assertThat(eventType, is(""));
+        }
+        {
+            String eventType = eventTypeExtractor.extractEventSubtype("{}");
+            assertThat(eventType, is(""));
+        }
+        {
+            String eventType = eventTypeExtractor.extractEventSubtype("{\"event\": {}}");
+            assertThat(eventType, is(""));
+        }
+    }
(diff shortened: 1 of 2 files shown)


slack-app-backend/src/test/java/test_locally/app_backend/events/EventTypeDetectorImplTest.java
```diff
@@ -11,6 +11,26 @@
 
     private EventTypeExtractor eventTypeExtractor = new EventTypeExtractorImpl();
 
+    @Test
+    public void notFound() {
+        {
+            String eventType = eventTypeExtractor.extractEventType("{}");
+            assertThat(eventType, is(""));
```
seratch, line 18: To verify the backward-compatiibilty

PR #492, opened 2020-06-17
Fix #491 Append OAuth redirect URI to Slack OAuth URL

### LINKED ISSUE #491: Allow users to specify redirect_uri during OAuth flow?
### Issue Type

- [ ] Bug
- [x] Enhancement / Feature request
- [x] Question
- [ ] Documentation

### Description

I have been using your excellent SDK to do two things:
1. Support installing an app
2. Support "Sign in with Slack"

My use case treats those as separate concerns in how my application works. My solution for supporting logging in with Slack OAuth involves defining 2 separate `com.slack.api.bolt.App` instances, with slightly different configuration, so that I can handle OAuth URLs that are distinct for each use case. One `App` instance is for installing the app, and one is for signing in with Slack.

This strategy has been mostly very practical; I don't have to write any custom OAuth logic since your SDK handles the important parts of the flow. I do need to replace a few components with subclasses to modify behavior slightly for my (perhaps unusual) approach here.
- I've subclassed `FileOAuthStateService` so that I can name the session cookies different things for either Slack `App`
- I need to subclass `App` itself in order to append a `redirect_uri` as a query param to the URL used to exchange an OAuth code for an access token 

The method `com.slack.api.bolt.App::getOauthInstallationUrl` builds URLs for exchanging an OAuth code for an access token. Slack's API documentation suggests that if there are multiple possible OAuth redirect URIs configured for an app, then using the `redirect_uri` query param helps Slack know which OAuth callback URI to send the user to after they accept the scopes.

**Here's my question**: would it make sense to include the redirect URI as configured on the `AppConfig` object when building that "OAuth Installation URL"? Defaulting to whatever the user configured on the `AppConfig` for their app seems sensible for a couple reasons:
- If someone is trying to do something like what I'm after here, they will likely have multiple Slack `App` instances, and the configuration will likely vary in terms of the OAuth URIs
- Not being explicit about which redirect URI to use seems potentially risky, because a user might reconfigure their app, and Slack seems to pick the first option in the list unless it's been explicitly told to use one or the other

For now, I've solved my own problem, but I'm wondering whether a general solution might be useful for others.

My proposal would effect this code: https://github.com/slackapi/java-slack-sdk/blob/master/bolt/src/main/java/com/slack/api/bolt/App.java#L317

```java
    public String getOauthInstallationUrl(String state) {
        AppConfig config = config();
        if (config.getClientId() == null || config.getScope() == null || state == null) {
            return null;
        } else {
            String scope = config.getScope() == null ? "" : config.getScope();
            if (config.isClassicAppPermissionsEnabled()) {
                // https://api.slack.com/authentication/migration
                return "https://slack.com/oauth/authorize" +
                        "?client_id=" + config.getClientId() +
                        "&scope=" + scope +
                        "&state=" + state +
						"&redirect_uri=" + config.getRedirectUri(); // Change here
            } else {
                String userScope = config.getUserScope() == null ? "" : config.getUserScope();
                return "https://slack.com/oauth/v2/authorize" +
                        "?client_id=" + config.getClientId() +
                        "&scope=" + scope +
                        "&user_scope=" + userScope +
                        "&state=" + state +
						"&redirect_uri=" + config.getRedirectUri(); // Change here
            }
        }
    }
```

---

### Requirements

* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/java-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've searched for any related issues and avoided creating a duplicate issue.

timrourke:
###  Summary

This PR appends a `redirect_uri` query param value based on the redirect URI specified in a given `AppConfig`, if such a value is set. This change enables Bolt application developers who need to configure their Slack applications with multiple OAuth redirect URIs to be explicit about where the OAuth callback should return to their Bolt application.

### Requirements

### DIFF
--- a/bolt/src/main/java/com/slack/api/bolt/App.java
+++ b/bolt/src/main/java/com/slack/api/bolt/App.java
@@ -36,8 +36,11 @@ import lombok.Data;
 import lombok.extern.slf4j.Slf4j;
 
 import java.io.IOException;
+import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.util.*;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.regex.Pattern;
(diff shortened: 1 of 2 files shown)


timrourke:
@seratch this PR should address https://github.com/slackapi/java-slack-sdk/issues/491. Let me know if I can help annotate this PR to keep things organized on your end.

bolt/src/main/java/com/slack/api/bolt/App.java
```diff
@@ -314,26 +317,60 @@ private void initOAuthServicesIfNecessary() {
...
+        }
+
+        try {
+            String urlEncodedRedirectUri = URLEncoder.encode(
+                appConfig.getRedirectUri(),
+                StandardCharsets.UTF_8.name()
+            );
+
+            return String.format("&redirect_uri=%s", urlEncodedRedirectUri);
+        } catch (UnsupportedEncodingException e) {
```
timrourke, line 372: `URLEncoder.encode` is virtually guaranteed to not throw because I am using a built-in encoding charset for URL-encoding the value for the `redirect_uri` query parameter. However, I'm open to suggestions for clarifying this a bit, or at least making this exception visible through logging. Throwing this exception seems inappropriate from this call site because it would require exception handling all the way up the stack in away that would change the public API of `App`.

seratch: > suggestions for clarifying this a bit, or at least making this exception visible through logging.  

As this is a well-known thing in Java and the exception is never thrown for sure, I don't see any issues with the current code.

bolt/src/test/java/test_locally/AppTest.java
```diff
@@ -51,6 +65,20 @@ public void getOauthInstallationUrl_v2() {
         assertEquals("https://slack.com/oauth/v2/authorize?client_id=123&scope=commands,chat:write&user_scope=search:read&state=state-value", url);
     }
 
+    @Test
+    public void getOauthInstallationUrl_v2_with_redirect_uri() {
```
seratch, line 69: Thanks for having corresponding unit tests in the right place!

PR #502, opened 2020-06-24
Fix #468 by implementing short-time cache for authorization middleware

### LINKED ISSUE #468: auth.test short-time cache for SingleTeamAuthorization
### Issue Type

- [?] Bug
- [ ] Enhancement / Feature request
- [x] Question
- [ ] Documentation

### Description

I am building a slack app with Bolt, and I'm observing a request to `auth.test` for every incoming request. This comes from here: https://github.com/slackapi/java-slack-sdk/blob/ac2a0107f1a2ac20db20704ac298a44aa4d35a28/bolt/src/main/java/com/slack/api/bolt/middleware/builtin/SingleTeamAuthorization.java#L37

The only parameter to the call (config.getSingleTeamBotToken()) is constant, so doing this for every incoming request seems a bit excessive, even though the Slack API docs state you can use the method a lot.

Should the response from `auth.test` be cached for a while instead of executed on every request?

seratch:
###  Summary

This pull request fixes #468 by introducing new options as below.

* AppConfig#authTestCacheEnabled (default: false)
* AppConfig#authTestCacheExpirationMillis (default: 3000)

The default behavior won't be changed. Only when a Bolt app turns the flag on, the cache will be enabled. 

The cache layer doesn't support distributed cache implementations (e.g., the ones using Memcached, Redis) by design. As mentioned in #468, the purpose of this cache is to reduce the number of `auth.test` API calls in Bolt apps that tend to receive lots of incoming requests from Slack in a short time of period.

TODO: Update document when releasing v1.1

### DIFF
--- a/bolt/src/main/java/com/slack/api/bolt/middleware/builtin/MultiTeamsAuthorization.java
+++ b/bolt/src/main/java/com/slack/api/bolt/middleware/builtin/MultiTeamsAuthorization.java
@@ -15,14 +15,19 @@ import com.slack.api.methods.MethodsClient;
 import com.slack.api.methods.SlackApiException;
 import com.slack.api.methods.response.auth.AuthTestResponse;
 import com.slack.api.model.block.LayoutBlock;
+import com.slack.api.util.thread.ExecutorServiceFactory;
 import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.extern.slf4j.Slf4j;
 
 import java.io.IOException;
 import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 
 import static com.slack.api.bolt.middleware.MiddlewareOps.isNoAuthRequiredRequest;
 import static com.slack.api.bolt.response.ResponseTypes.ephemeral;
@@ -45,6 +50,7 @@ public class MultiTeamsAuthorization implements Middleware {
 
     // token -> auth.test response
     private final ConcurrentMap<String, CachedAuthTestResponse> tokenToAuthTestCache = new ConcurrentHashMap<>();
+    private final Optional<ScheduledExecutorService> tokenToAuthTestCacheCleaner;
 
     private boolean alwaysRequestUserTokenNeeded;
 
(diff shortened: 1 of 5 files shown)


seratch:
Will fix the build failure. It’s just a matter of test code.

bolt/src/main/java/com/slack/api/bolt/middleware/builtin/MultiTeamsAuthorization.java
```diff
@@ -31,6 +36,16 @@
...
 
+    @Data
+    @AllArgsConstructor
+    static class CachedAuthTestResponse {
+        private AuthTestResponse response;
+        private long cachedMillis;
+    }
+
+    // token -> auth.test response
+    private final ConcurrentMap<String, CachedAuthTestResponse> tokenToAuthTestCache = new ConcurrentHashMap<>();
```
seratch, line 47: We may want to clean up very old cache data to avoid excessive memory usage in extreme cases. As I don't want to add any dependencies only for this, we can implement a simple mechanism on our own.

bolt/src/main/java/com/slack/api/bolt/middleware/builtin/SingleTeamAuthorization.java
```diff
@@ -22,6 +28,9 @@
     private final AppConfig appConfig;
     private final InstallationService installationService;
 
+    private Optional<AuthTestResponse> cachedAuthTestResponse = Optional.empty();
+    private AtomicLong lastCachedMillis = new AtomicLong(0L);
```
seratch, line 32: Thanks to @aoberoi 's suggestion here: https://github.com/slackapi/java-slack-sdk/issues/468#issuecomment-650254481, we can remove this TTL and store the response when booting a Bolt app as with Bolt for JS.

seratch: I came to think removing TTL from only single team auth may be confusing. I decided to add a "permanent" option by giving a negative TTL value instead.

bolt/src/main/java/com/slack/api/bolt/middleware/builtin/MultiTeamsAuthorization.java
seratch: done

seratch:
This PR is ready for review. I'll wait for other maintianers' responses for a few business days.

PR #577, opened 2020-10-05
Change the fallback behavior of Events API handlers to be compatible with Bolt JS/Python

seratch:
While doing tests with Workflow Steps, I found the current fallback logic of Events API handlers in Bolt for Java is to ignore unknown events by returning 200 OK. While it may be convenient for some apps, this is obviously incompatible with other Bolt frameworks. So, since v1.2, we will change this behavior.

## Requirements

Please read the [Contributing guidelines](https://github.com/slackapi/java-slack-sdk/blob/main/.github/contributing.md) and [Code of Conduct](https://slackhq.github.io/code-of-conduct) before creating this issue or pull request. By submitting, you are agreeing to those rules.

### DIFF
--- a/bolt-aws-lambda/src/test/java/test_locally/Issue419UseCaseTest.java
+++ b/bolt-aws-lambda/src/test/java/test_locally/Issue419UseCaseTest.java
@@ -79,6 +79,7 @@ public class Issue419UseCaseTest {
                 headers.put("additional-header", Arrays.asList("foo"));
                 return Response.builder().statusCode(200).headers(headers).build();
             });
+            app.message("Hello", (req, ctx) -> ctx.ack());
 
             SampleHandler handler = new SampleHandler(app);
 
--- a/bolt/src/main/java/com/slack/api/bolt/App.java
+++ b/bolt/src/main/java/com/slack/api/bolt/App.java
@@ -930,16 +930,16 @@ public class App {
                 case Event: {
                     if (eventsDispatcher.isRunning()) {
                         eventsDispatcher.enqueue(slackRequest.getRequestBodyAsString());
+                        return Response.ok();
                     }
                     EventRequest request = (EventRequest) slackRequest;
                     BoltEventHandler<Event> handler = eventHandlers.get(request.getEventTypeAndSubtype());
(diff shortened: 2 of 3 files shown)


bolt-aws-lambda/src/test/java/test_locally/Issue419UseCaseTest.java
```diff
@@ -79,6 +79,7 @@ public void enableAdditionalListenersRuntime_issue_419() throws Exception {
                 headers.put("additional-header", Arrays.asList("foo"));
                 return Response.builder().statusCode(200).headers(headers).build();
             });
+            app.message("Hello", (req, ctx) -> ctx.ack());
```
seratch, line 82: Due to the current behavior, this test has been passing without this required handler.

bolt/src/main/java/com/slack/api/bolt/App.java
```diff
@@ -930,16 +930,16 @@ protected Response runHandler(Request slackRequest) throws IOException, SlackApi
                 case Event: {
                     if (eventsDispatcher.isRunning()) {
                         eventsDispatcher.enqueue(slackRequest.getRequestBodyAsString());
+                        return Response.ok();
```
seratch, line 933: This is a primitive event dispatcher in slack-app-backend module. Although most developers probably don't use this in Bolt for Java, Bolt enables using it in Bolt apps. When developers use this, event requests should be still acknowledged anyway. So, I added the return statement here.

bolt/src/test/java/test_locally/app/MessageTest.java
```diff
@@ -219,11 +222,12 @@ public void bot_skipped() throws Exception {
             return ctx.ack();
         });
 
+        // "This is a message sent by a bot user."
         EventsApiPayload<MessageBotEvent> payload = buildUserBotMessagePayload();
 
         EventRequest req = buildRequest(gson.toJson(payload));
         Response response = app.run(req);
-        assertEquals(200L, response.getStatusCode().longValue());
```
seratch, line 226: This has been just a wrong assertion.

PR #598, opened 2020-10-26
Update BoltLibraryVersion.java : Runtime of method is greater than accessing variable

dev-segal:
(Describe the goal of this PR. Mention any related Issue numbers)
The process involved with a method (time for one) is greater overall than accessing a variable, thus the change.

## Requirements

### DIFF
--- a/bolt/src/main/java/com/slack/api/bolt/meta/BoltLibraryVersion.java
+++ b/bolt/src/main/java/com/slack/api/bolt/meta/BoltLibraryVersion.java
@@ -1,11 +1,9 @@
 package com.slack.api.bolt.meta;
 
 public final class BoltLibraryVersion {
+    public static final String VERSION = "1.3.0-SNAPSHOT";
+    
     private BoltLibraryVersion() {
-    }
-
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for the suggestion! I think it's not a large overhead as Java VM runtime may optimize this type of simple calls. That being said, I think this may be a good improvement.

However, this is a breaking change to existing public APIs so that we don't want to remove the `get()` method in the short term. We can revisit this in v2.0 (we haven't decided when to release it yet).

If you see the necessity to reduce the `get()` calls inside this framework, I'm happy to change the internals to have call result cache. But the calls happen only initializing `AppConfig` instance. This means there is no overhead affecting Bolt app's response time.

dev-segal:
Yes, absolutely agreed. Perhaps using it in v2+ perhaps could be minutely better, but it's definitely not an issue of enormous magnitude, thanks for looking it over.

seratch:
As we are not going to apply this change in the short term, please allow me to close this issue now. Thanks for taking the time to make this PR!

PR #677, opened 2021-01-28
Fix #676 Add an easier way to access Events API retry info

### LINKED ISSUE #676: Add an easier way to access X-Slack-Retry-Num / X-Slack-Retry-Reason in Bolt apps
We can add X-Slack-Retry-Num and X-Slack-Retry-Reason values to `com.slack.api.bolt.context.builtin.EventContext` class.

---

~jayyhkwon Ah ... you are right. I have to say there is no way to directly access request headers in listeners. I will take a look at possible ways to support this use case in the future versions. As a workaround, you can use a global middleware this way:

```java
app.use((req, resp, chain) -> {
  String retryNum = req.getHeaders().getFirstValue("X-Slack-Retry-Num");
  req.getContext().getAdditionalValues().put("x-slack-retry-num", retryNum);
  return chain.next(req);
});
app.event(AppMentionEvent.class, (req, ctx) -> {
  String retryNum = ctx.getAdditionalValues().get("x-slack-retry-num");
  return ctx.ack();
});
```

_Originally posted by @seratch in https://github.com/slackapi/java-slack-sdk/issues/646#issuecomment-745073746_

seratch:
This pull request enables Events API listeners to access x-slack-retry-num / x-slack-retry-reason header values (retry_attempt / retry_reason in Socket Mode) in the following way.

```java
app.event(AppMentionEvent.class, (req, ctx) -> {
  Integer num = ctx.getRetryNum(); // 0-3
  String reason = ctx.getRetryReason(); // timeout, http_error etc
  return ctx.ack();
});
```

Fixes #676 

* [x] **bolt-socket-mode** (Bolt for Java - optional modules)

## Requirements

### DIFF
--- a/bolt/src/main/java/com/slack/api/bolt/context/builtin/EventContext.java
+++ b/bolt/src/main/java/com/slack/api/bolt/context/builtin/EventContext.java
@@ -15,4 +15,11 @@ public class EventContext extends Context implements SayUtility {
 
     private String channelId;
 
+    // X-Slack-Retry-Num: 2 in HTTP Mode
+    // "retry_attempt": 0, in Socket Mode
+    private Integer retryNum;
+    // X-Slack-Retry-Reason: http_error in HTTP Mode
+    // "retry_reason": "timeout", in Socket Mode
+    private String retryReason;
+
 }
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -10,7 +10,8 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    # ubuntu-20.04 may be a bit unstable for this build
+    runs-on: ubuntu-18.04
     timeout-minutes: 10
     strategy:
       matrix:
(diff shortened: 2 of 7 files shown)


bolt/src/main/java/com/slack/api/bolt/context/builtin/EventContext.java
```diff
@@ -15,4 +15,11 @@
 
     private String channelId;
 
+    // X-Slack-Retry-Num: 2 in HTTP Mode
+    // "retry_attempt": 0, in Socket Mode
```
seratch, line 19: @stevengill @mwbrooks 
The names are a bit different here. Do we prefer `retryAttempt`? I personally feel `retryNum` is fine as it's a number.

mwbrooks: I'd personally prefer to keep the names consistent between the headers and variables. It makes it easier to understand that the two are identical and intended to match. So I'd lean toward `retryNum` as well.

mwbrooks (approved):

sagar-ranglani:
Any plans on adding this for boltjs?

PR #690, opened 2021-02-12
Fix #689 by adding flags in AppConfig

### LINKED ISSUE #689: Allow removing default middleware in Bolt
Firstly, thank you for building this SDK -- it's awesome!

Currently, you can call the [`use`](https://github.com/slackapi/java-slack-sdk/blob/d902f86f699eb9b03f3f881942af159b39b90ac2/bolt/src/main/java/com/slack/api/bolt/App.java#L519) method to register middleware on the `App` class, but you cannot modify the list. For example, in my case I wanted to remove the `RequestVerification` middleware because we have a proxy in front of our server that already checks the Slack signatures, but I was forced to sub-class `App` in order to override [`buildDefaultMiddlewareList`](https://github.com/slackapi/java-slack-sdk/blob/d902f86f699eb9b03f3f881942af159b39b90ac2/bolt/src/main/java/com/slack/api/bolt/App.java#L93) to filter out that middleware.

It'd be nice if there was a way to modify the middleware list outside of `App`'s constructor, or perhaps just expose `buildDefaultMiddlewareList` as a `public static` method, so you could do something like this:
```kotlin
val middleware = App.buildDefaultMiddlewareList().filter { it !is RequestVerification }
val app = App(appConfig, middleware)
```

Alternatively, for my particular use case, a configuration flag in `AppConfig` controlling whether that specific middleware gets included in the list would have done the trick as well.

Cc: @gal-yardeni

seratch:
This pull request fixes #689 by adding flags in `AppConfig`. Refer to the issue for the context. 

We do not usually recommend turning the built-in middleware off. Especially, disabling request verification just for easiness is not great from security perspective. These flags are available only for the use case like #689 or debugging purposes.

## Requirements


luispollo:
@seratch Thank you for picking this up so quickly! ❤️ 

The PR looks good to me! My only suggestion would be to update [the middleware documentation](https://github.com/slackapi/java-slack-sdk/blob/3d4b518d83de40661a0e21b317e8cdcbe7ab14d2/docs/guides/bolt-basics.md#middleware) to add a note about the ability to disable default middleware, which would be a good place to add a warning about the fact that disabling request verification is not recommended for security reasons, as you mentioned in the description.

seratch:
@luispollo Thanks for the suggestion! I've updated the documents in this pull request ✅

docs/guides/bolt-basics.md
```diff
@@ -272,6 +272,31 @@ A set of the built-in middleware precedes your custom middleware. So, if the app
 
 The most common would be the case where a request has been denied by **RequestVerification** middleware. After the denial, any middleware won't be executed, so that the above middleware also doesn't work for the case.
 
+#### Customize the Built-in Middleware List
+
+Bolt turns the following middleware on by default.
```
mwbrooks, line 277:
```suggestion
Bolt turns the following middleware on by default:
```

docs/guides/bolt-basics.md
```diff
@@ -272,6 +272,31 @@ A set of the built-in middleware precedes your custom middleware. So, if the app
...
+* [SingleTeamAuthorization](https://github.com/slackapi/java-slack-sdk/blob/main/bolt/src/main/java/com/slack/api/bolt/middleware/builtin/SingleTeamAuthorization.java) or [MultiTeamsAuthorization](https://github.com/slackapi/java-slack-sdk/blob/main/bolt/src/main/java/com/slack/api/bolt/middleware/builtin/MultiTeamsAuthorization.java) to lookup the valid OAuth access token for handling a request
+* [IgnoringSelfEvents](https://github.com/slackapi/java-slack-sdk/blob/main/bolt/src/main/java/com/slack/api/bolt/middleware/builtin/IgnoringSelfEvents.java) to skip the events generated by this app's bot user (this is useful for avoiding code error causing an infinite loop)
+* [SSLCheck](https://github.com/slackapi/java-slack-sdk/blob/main/bolt/src/main/java/com/slack/api/bolt/middleware/builtin/SSLCheck.java) to handle `ssl_check=1` requests from Slack
+
+Although generally do not recommend disabling them as they are commonly necessary, you can disable them using the flags like `ignoringSelfEventsEnabled` in `AppConfig` objects.
```
mwbrooks, line 284:
```suggestion
Although we generally do not recommend disabling these middleware as they are commonly necessary, you can disable them using the flags like `ignoringSelfEventsEnabled` in `AppConfig` objects.
```

mwbrooks (approved):
This is a great addition. Is it something we want to bring to JavaScript and Python?

seratch:
>Is it something we want to bring to JavaScript and Python?

It may be worth considering to add these options to Python as its `App` has a quite similar structure. As for JavaScript, request verification is HTTPReceiver / ExpressReceiver's responsibility. Thus, if we add this, the options should be available in these receivers' constructors.

seratch:
Thanks for the suggestions on the document 🙇

gal-yardeni:
Hi @seratch, thank you for fixing the issue for us! When it will be available for use at the SDK level?

seratch:
@gal-yardeni We'll release a new version within a few hours!

gal-yardeni:
that's awesome! thanks for the update!

PR #703, opened 2021-03-04
Add built-in tokens_revoked/app_uninstalled event handlers

seratch:
As of today, developers have to implement their own logic to clean up database like this: https://github.com/seratch/send-it-later-for-slack/blob/e5f3555f347826576099f2f8551b03089ba7479d/app/listeners.py#L209-L226 

I came to think that Bolt should provide better abstraction of installation store operations for it. If other maintainers are fine with this approach, I'm happy to add the same functionalities to other Bolt frameworks.

This pull request adds built-in event handlers for token revocations and app uninstallation. The handy way to turn the additional features on is to use `enableTokenRevocationHandlers()` method as below:

```java
var app = new App().enableTokenRevocationHandlers();
```

This is equivalent to the following code:

```java
var app = new App();
app.event(TokensRevokedEvent.class, app.defaultTokensRevokedEventHandler());
app.event(AppUninstalledEvent.class, app.defaultAppUninstalledEventHandler());
```

## Requirements

### DIFF
--- a/bolt/src/main/java/com/slack/api/bolt/handler/builtin/DefaultAppUninstalledEventHandler.java
+++ b/bolt/src/main/java/com/slack/api/bolt/handler/builtin/DefaultAppUninstalledEventHandler.java
@@ -1,0 +1,55 @@
+package com.slack.api.bolt.handler.builtin;
+
+import com.slack.api.app_backend.events.payload.EventsApiPayload;
+import com.slack.api.bolt.context.builtin.EventContext;
+import com.slack.api.bolt.handler.BoltEventHandler;
+import com.slack.api.bolt.response.Response;
+import com.slack.api.bolt.service.InstallationService;
+import com.slack.api.methods.SlackApiException;
+import com.slack.api.model.event.AppUninstalledEvent;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.concurrent.ExecutorService;
+
+public class DefaultAppUninstalledEventHandler implements BoltEventHandler<AppUninstalledEvent> {
+
+    private final InstallationService installationService;
+    private final ExecutorService executorService;
+    private final Logger logger = LoggerFactory.getLogger(DefaultAppUninstalledEventHandler.class);
+
+    public static DefaultAppUninstalledEventHandler getInstance(
+            InstallationService installationService,
+            ExecutorService executorService) {
+        return new DefaultAppUninstalledEventHandler(installationService, executorService);
+    }
+
+    public DefaultAppUninstalledEventHandler(
+            InstallationService installationService,
+            ExecutorService executorService) {
+        this.installationService = installationService;
+        this.executorService = executorService;
+    }
+
+    @Override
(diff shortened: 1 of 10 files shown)


bolt-servlet/src/test/java/samples/OAuthSample.java
```diff
@@ -9,16 +12,36 @@
 public class OAuthSample {
 
     public static void main(String[] args) throws Exception {
-        App app = new App().asOAuthApp(true);
+        AppConfig config = ResourceLoader.loadAppConfig("appConfig_oauth.json");
+        App app = new App(config).asOAuthApp(true)
+                // Enable built-in tokens_revoked / app_uninstalled event handlers
+                .enableTokenRevocationHandlers();
```
seratch, line 18: Any suggestions to this naming are welcomed!

bolt/src/main/java/com/slack/api/bolt/handler/builtin/DefaultAppUninstalledEventHandler.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Override
+    public Response apply(
+            EventsApiPayload<AppUninstalledEvent> payload,
+            EventContext context) throws IOException, SlackApiException {
+        // NOTE: this payload does not have authorizations[]
+        String enterpriseId = payload.getEnterpriseId();
+        String teamId = payload.getTeamId();
+
+        this.executorService.submit(() -> {
```
seratch, line 43: This is the recommended way to run async operations in Bolt for Java.

bolt/src/main/java/com/slack/api/bolt/service/InstallationService.java
```diff
@@ -60,4 +60,12 @@ default String getInstallationGuideText(String enterpriseId, String teamId, Stri
         return null;
     }
 
+    /**
+     * Deletes all installation data for given workspace or organization.
+     */
+    default void deleteAll(String enterpriseId, String teamId) {
```
seratch, line 66: We don't require to implement this method as long as developers don't use the built-in `app_uninstalled` event handler.

bolt-servlet/src/test/java/samples/OAuthSample.java
```diff
@@ -9,16 +12,36 @@
...
+
+        // NOTE: the following also works in the same way:
+        // app.event(TokensRevokedEvent.class, app.defaultTokensRevokedEventHandler());
+        // app.event(AppUninstalledEvent.class, app.defaultAppUninstalledEventHandler());
```
mwbrooks, line 22: I really like that each one has been separated out!

bolt-servlet/src/test/java/samples/OAuthSample.java
mwbrooks: It's a tricky name, but I don't have any better suggestions. It made sense to me, so I'm okay with it!

mwbrooks (approved):

seratch:
@mwbrooks Thanks for your review!

Issue #749, opened 2021-05-31
[slack-api-client] NullPointerException in runtime if proxy port is empty


alexkvak:
#### The Slack SDK version

```
[INFO] +- com.slack.api:slack-api-client:jar:1.8.0:compile
[INFO] |  +- com.slack.api:slack-api-model:jar:1.8.0:compile
```

#### Java Runtime version

```
java version "11.0.4" 2019-07-16 LTS
Java(TM) SE Runtime Environment 18.9 (build 11.0.4+10-LTS)
Java HotSpot(TM) 64-Bit Server VM 18.9 (build 11.0.4+10-LTS, mixed mode)
```

#### OS info

```
ProductName:    macOS
ProductVersion: 11.4
BuildVersion:   20F71
Darwin Kernel Version 20.5.0: Sat May  8 05:10:33 PDT 2021; root:xnu-7195.121.3~9/RELEASE_X86_64
```

#### Steps to reproduce:
The following code throws `NullPointerException`

```java
SlackConfig config = new SlackConfig();
config.setProxyUrl("http://proxy.com");
Slack.getInstance(config);
```

while `java.net.InetSocketAddress` constructor expects `int` as second param but got `null` in https://github.com/slackapi/java-slack-sdk/blob/main/slack-api-client/src/main/java/com/slack/api/util/http/SlackHttpClient.java#L59

#### Expected result:

`0` should be passed as second param to `java.net.InetSocketAddress` constructor

#### Actual result:

`NullPointerException` throws

seratch:
Hi @alexkvak, thank you very much for taking the time to report this issue! 

This seems to be an obvious bug. We will fix this issue in the next patch version. 

By any chance, if you are interested in working on the fix, we're so happy to have your name in [the contributor list](https://github.com/slackapi/java-slack-sdk/graphs/contributors) 🙌

abhikarkamkar:
@seratch can I pick this up

seratch:
@alexkvak I wanted to know the reason you suggested zero as the expected port in this scenario.

>0 should be passed as second param to java.net.InetSocketAddress constructor

According to the class's javadoc,

>A valid port value is between 0 and 65535. A port number of zero will let the system pick up an ephemeral port in a bind operation.

With the zero port, the proxy URL will be `{host}:{ephemeral port}`. Is this the expected behavior for common patterns?
I'm thinking that 80 may be more suitable for covering common patterns. Am I missing something?

[...]

PR #751, opened 2021-06-02
Fix for [slack-api-client] NullPointerException in runtime if proxy port is empty

abhikarkamkar:
## Fix [NullPointerException in runtime if proxy port is empty] (https://github.com/slackapi/java-slack-sdk/issues/749)

currently null is passed to java.net.InetSocketAddress constructor which raises NullPointerException
to overcome this 0 is passed as second param to java.net.InetSocketAddress constructor

### DIFF
--- a/slack-api-client/src/main/java/com/slack/api/util/http/ProxyUrlUtil.java
+++ b/slack-api-client/src/main/java/com/slack/api/util/http/ProxyUrlUtil.java
@@ -48,14 +48,14 @@ public class ProxyUrlUtil {
                     .username(userAndPassword[0])
                     .password(userAndPassword[1])
                     .host(hostAndPort[0])
-                    .port(hostAndPort.length == 2 ? Integer.parseInt(hostAndPort[1]) : 0)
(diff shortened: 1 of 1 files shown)


slack-api-client/src/main/java/com/slack/api/util/http/ProxyUrlUtil.java
```diff
@@ -48,14 +48,14 @@ public static ProxyUrl parse(String proxyUrl) {
                     .username(userAndPassword[0])
                     .password(userAndPassword[1])
                     .host(hostAndPort[0])
-                    .port(hostAndPort.length == 2 ? Integer.valueOf(hostAndPort[1]) : null)
+                    .port(hostAndPort.length == 2 ? Integer.parseInt(hostAndPort[1]) : 80)
```
seratch, line 51: I like the slight improvement to use `Integer.parseInt` here.

seratch (approved):
LGTM 👍 

As you know, I'm asking the issue reporter a question in the issue. just in case, I will hold off merging this for a few days.

PR #832, opened 2021-09-01
Remove cached slack, client in App to reflect changes in AppConfig

seratch:
This pull request improves the internals of `App` class as below:

* `App#getSlack()` always returns a `Slack` instance with the latest configuration
* `App#getClient()` always returns a `MethodsClient` instance with the latest configuration
* Fix potential NPE error patterns

## Requirements

### DIFF
--- a/bolt/src/main/java/com/slack/api/bolt/App.java
+++ b/bolt/src/main/java/com/slack/api/bolt/App.java
@@ -70,12 +70,16 @@ public class App {
     /**
      * Slack instance and its configurations to be used for API calls.
      */
-    private final Slack slack;
+    public Slack getSlack() {
+        return config().getSlack();
+    }
 
     /**
      * The default Web API client without any tokens.
      */
-    private final MethodsClient client;
+    public MethodsClient getClient() {
+        return config().getSlack().methods();
+    }
 
     /**
      * The built-in handy way to run operations asynchronously. It's totally fine to use your own one instead.
@@ -441,8 +445,6 @@ public class App {
 
     public App(AppConfig appConfig, Slack slack, List<Middleware> middlewareList) {
         this.appConfig = appConfig;
-        this.slack = slack;
-        this.client = this.slack.methods();
         this.executorService = ExecutorServiceFactory.createDaemonThreadPoolExecutor(
                 "bolt-app-threads",
                 this.appConfig.getThreadPoolSize());
@@ -482,11 +484,11 @@ public class App {
     }
 
     public Slack slack() {
-        return this.slack;
+        return getSlack();
     }
 
     public MethodsClient client() {
-        return this.client;
+        return getClient();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


bolt/src/main/java/com/slack/api/bolt/App.java
```diff
@@ -70,12 +70,16 @@
     /**
      * Slack instance and its configurations to be used for API calls.
      */
-    private final Slack slack;
```
seratch, line 73: Lombok's `@Data` annotation generates `getSlack()` method for this field.

bolt/src/main/java/com/slack/api/bolt/App.java
```diff
@@ -441,8 +445,6 @@ public App(AppConfig appConfig, List<Middleware> middlewareList) {
 
     public App(AppConfig appConfig, Slack slack, List<Middleware> middlewareList) {
         this.appConfig = appConfig;
-        this.slack = slack;
-        this.client = this.slack.methods();
```
seratch, line 445: Removed these cached instances. Accessing `config().getSlack()` instead has almost zero overhead.

slack-api-client/src/main/java/com/slack/api/methods/impl/TeamIdCache.java
```diff
@@ -32,15 +32,16 @@ public String lookupOrResolve(String token) {
                 FormBody.Builder form = toForm(AuthTestRequest.builder().token(newToken).build());
                 Response response = methodsImpl.runPostFormWithToken(form, Methods.AUTH_TEST, token);
                 AuthTestResponse authTest = methodsImpl.parseJsonResponseAndRunListeners(null, null, response, AuthTestResponse.class);
-                if (authTest.isOk()) {
+                if (authTest != null && authTest.isOk()) {
```
seratch, line 35: While writing a mock API server and doing tests with it, I found the potential null pattern. Except the situation where the Slack server-side returns empty response body, `authTest` cannot be null, though.

slack-api-client/src/main/java/com/slack/api/methods/impl/TeamIdCache.java
```diff
@@ -32,15 +32,16 @@ public String lookupOrResolve(String token) {
...
                     if (log.isDebugEnabled()) {
                         log.debug("Created cache for an auth.test API call (token: {}, team_id: {})",
                                 token.substring(0, 16) + "...", authTest.getTeamId());
                     }
                     // for org admin user's token, this value can be an enterprise_id
                     return authTest.getTeamId();
                 } else {
-                    log.error("Got an unsuccessful response from auth.test API (error: {})", authTest.getError());
+                    String error = authTest != null ? authTest.getError() : "";
```
filmaj, line 43: Perhaps in the case that `authTest` is null, we set a more descriptive error than the empty string? I know you said `authTest` cannot be null - but who knows how this code will change in the future 😃 

```suggestion
                    String error = authTest != null ? authTest.getError() : "AuthTestResponse is null";
```

filmaj (approved):
👍

slack-api-client/src/main/java/com/slack/api/methods/impl/TeamIdCache.java
seratch:
```suggestion
                    String error = authTest != null ? authTest.getError() : "(empty response)";
```

seratch: Thanks! Perhaps, mentioning an empty response body would be easier to understand

PR #850, opened 2021-09-23
Add getHttpResponseHeaders() method to Web API responses

seratch:
This pull request adds two new methods to the Web API's response types.

* getHttpResponseHeaders
* setHttpResponseHeaders

These methods are useful for developers that would like to access any of the returned HTTP response headers of Web APIs. Particularly, the `x-oauth-scopes` header is useful to get the list of the associated OAuth scopes with the used OAuth token.

```java
  String token = System.getenv("SLACK_BOT_TOKEN");
  AuthTestResponse response = Slack.getInstance().methods().authTest(r -> r.token(token));
  // To support the situation where a header appears in a response several times, #get(key) returns List<String>
  String scopes = response.getHttpResponseHeaders().get("x-oauth-scopes").get(0);
  System.out.println(scopes); // app_mentions:read,chat:write,files:read ...
}
```

As I mentioned at https://github.com/slackapi/bolt-js/issues/1106#issuecomment-925443768, node-slack-sdk merges the `x-oauth-scopes` value into `response_metadata`.

In the Python SDK, [the response class](https://github.com/slackapi/python-slack-sdk/blob/v3.11.2/slack_sdk/web/slack_response.py#L61-L63) has the following properties:

* data (dict or str)
* headers (dict)
* status_code (int)

Also, the response class provides getter methods for the properties under the `data` (= both `response.data.ok` and `response.ok` work).

For this Java SDK, I prefer the Python's approach that does not modify the response body (=data in the response class).

## Requirements

### DIFF
--- a/slack-api-client/src/main/java/com/slack/api/methods/SlackApiErrorResponse.java
+++ b/slack-api-client/src/main/java/com/slack/api/methods/SlackApiErrorResponse.java
@@ -2,6 +2,9 @@ package com.slack.api.methods;
 
 import lombok.Data;
 
+import java.util.List;
+import java.util.Map;
+
 @Data
 public class SlackApiErrorResponse implements SlackApiTextResponse {
 
@@ -10,4 +13,5 @@ public class SlackApiErrorResponse implements SlackApiTextResponse {
     private String error;
     private String needed;
     private String provided;
+    private transient Map<String, List<String>> httpResponseHeaders;
 }
--- a/slack-api-client/src/main/java/com/slack/api/methods/SlackApiResponse.java
+++ b/slack-api-client/src/main/java/com/slack/api/methods/SlackApiResponse.java
@@ -1,4 +1,17 @@
 package com.slack.api.methods;
 
+import java.util.List;
+import java.util.Map;
+
 public interface SlackApiResponse {
+
+    /**
+     * Returns all the HTTP response headers in the API response. The keys are lower-cased.
+     */
+    Map<String, List<String>> getHttpResponseHeaders();
+
+    /**
+     * Sets the response headers. Pass a Map object with lower-cased keys.
+     */
+    void setHttpResponseHeaders(Map<String, List<String>> headers);
 }
--- a/slack-api-client/src/main/java/com/slack/api/methods/impl/MethodsClientImpl.java
+++ b/slack-api-client/src/main/java/com/slack/api/methods/impl/MethodsClientImpl.java
@@ -221,7 +221,7 @@ import lombok.extern.slf4j.Slf4j;
 import okhttp3.*;
 
 import java.io.IOException;
-import java.util.Optional;
+import java.util.*;
 
 import static com.slack.api.methods.RequestFormBuilder.toForm;
 import static com.slack.api.methods.RequestFormBuilder.toMultipartBody;
@@ -2934,7 +2934,10 @@ public class MethodsClientImpl implements MethodsClient {
             if (response.isSuccessful()) {
                 try {
                     Gson gson = GsonFactory.createSnakeCase(slackHttpClient.getConfig());
-                    return gson.fromJson(body, AdminAnalyticsGetFileResponse.class);
+                    AdminAnalyticsGetFileResponse apiResponse =
+                            gson.fromJson(body, AdminAnalyticsGetFileResponse.class);
+                    apiResponse.setHttpResponseHeaders(toLowerCasedKeyMap(response.headers()));
+                    return apiResponse;
                 } finally {
                     slackHttpClient.runHttpResponseListeners(response, body);
                 }
@@ -2947,6 +2950,7 @@ public class MethodsClientImpl implements MethodsClient {
             AdminAnalyticsGetFileResponse apiResponse = new AdminAnalyticsGetFileResponse();
             apiResponse.setOk(true);
             apiResponse.setFileStream(response.body().byteStream());
+            apiResponse.setHttpResponseHeaders(toLowerCasedKeyMap(response.headers()));
             return apiResponse;
         }
     }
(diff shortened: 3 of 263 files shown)


slack-api-client/src/main/java/com/slack/api/methods/SlackApiErrorResponse.java
```diff
@@ -10,4 +13,5 @@
     private String error;
     private String needed;
     private String provided;
+    private transient Map<String, List<String>> httpResponseHeaders;
```
seratch, line 16: Having transient tells the JSON library that this property should not be included in serialized JSON string data. No side effect for other use cases.

slack-api-client/src/main/java/com/slack/api/methods/SlackApiResponse.java
```diff
@@ -1,4 +1,17 @@
 package com.slack.api.methods;
 
+import java.util.List;
+import java.util.Map;
+
 public interface SlackApiResponse {
+
+    /**
+     * Returns all the HTTP response headers in the API response. The keys are lower-cased.
```
seratch, line 9: MethodsClient ensures all the keys are lower-cased.

slack-api-client/src/main/java/com/slack/api/methods/impl/MethodsClientImpl.java
```diff
@@ -2934,7 +2934,10 @@ public AdminAnalyticsGetFileResponse adminAnalyticsGetFile(AdminAnalyticsGetFile
             if (response.isSuccessful()) {
                 try {
                     Gson gson = GsonFactory.createSnakeCase(slackHttpClient.getConfig());
-                    return gson.fromJson(body, AdminAnalyticsGetFileResponse.class);
+                    AdminAnalyticsGetFileResponse apiResponse =
+                            gson.fromJson(body, AdminAnalyticsGetFileResponse.class);
+                    apiResponse.setHttpResponseHeaders(toLowerCasedKeyMap(response.headers()));
```
seratch, line 2939: Only this method is different from others because the endpoint can return either text body or binary data. Thus, I repeated the same logic.

slack-api-client/src/main/java/com/slack/api/methods/impl/MethodsClientImpl.java
```diff
@@ -2947,6 +2950,7 @@ public AdminAnalyticsGetFileResponse adminAnalyticsGetFile(AdminAnalyticsGetFile
             AdminAnalyticsGetFileResponse apiResponse = new AdminAnalyticsGetFileResponse();
             apiResponse.setOk(true);
             apiResponse.setFileStream(response.body().byteStream());
+            apiResponse.setHttpResponseHeaders(toLowerCasedKeyMap(response.headers()));
```
seratch, line 2953: same as above

slack-api-client/src/main/java/com/slack/api/methods/impl/MethodsClientImpl.java
```diff
@@ -3129,6 +3133,7 @@ protected String getToken(SlackApiRequest request) {
                         metricsDatastore.incrementUnsuccessfulCalls(executorName, teamId, methodName);
                     }
                 }
+                apiResponse.setHttpResponseHeaders(toLowerCasedKeyMap(response.headers()));
```
seratch, line 3136: This part appends the headers to the API response object.

slack-api-client/src/test/java/test_with_remote_apis/methods/auth_Test.java
```diff
@@ -216,4 +216,12 @@ public void authTeamsList_org_level_app() throws Exception {
         }
     }
 
+    @Test
+    public void authTest_x_oauth_scopes() throws IOException, SlackApiException {
+        AuthTestResponse response = slack.methods().authTest(req -> req.token(botToken));
+        assertThat(response.getError(), is(nullValue()));
+        String scopes = response.getHttpResponseHeaders().get("x-oauth-scopes").get(0);
```
seratch, line 223: Added a few tests like this in this PR

slack-api-client/src/main/java/com/slack/api/methods/SlackApiResponse.java
```diff
@@ -1,4 +1,17 @@
...
+     */
+    Map<String, List<String>> getHttpResponseHeaders();
```
seratch, line 11: To support the situation where a header appears in a response several times, I use `List<String>` for the values. I'm sure that this is a common design.

srajiang (approved):
This looks good to me! Interesting to know the difference in implementation between node-slack-sdk and python / java. For my learning - it looks like all headers will be available with these changes - other than `x-oauth-scopes` which has been requested by community, are there others the community has wanted to have access to?

slack-api-client/src/main/java/com/slack/api/methods/SlackApiResponse.java
filmaj: Interesting, I don't know if I've seen this approach before, but your approach makes sense to me.
Out of curiosity, are there situations the Slack backend returns multiple values for a single header?
Perhaps we can add a test to ensure multiple values for a single header are represented properly in this construct?

filmaj (approved):
Left a minor comment, otherwise LGTM! :shipit:

slack-api-client/src/main/java/com/slack/api/methods/SlackApiResponse.java
seratch: >Out of curiosity, are there situations the Slack backend returns multiple values for a single header?

No, there aren't at this moment but it may be in the future.

>Perhaps we can add a test to ensure multiple values for a single header are represented properly in this construct?

The okhttp library does this and we can trust its behavior but it'd be good to add some tests on our end.

PR #859, opened 2021-09-29
Fix #857 Add the missing "pronouns" field to User.Profile data class

### LINKED ISSUE #857: Add the missing "pronouns" field to User.Profile data class
With the addition of [pronouns](https://slack.com/intl/en-gb/help/articles/1500011115781-Manage-pronoun-display-for-member-profiles) to user's profiles, there's a field returned as part of the Profile structure ```pronouns```. This is returned as part of the response when using the [tester](https://api.slack.com/methods/users.profile.get/test) but it isn't represented in the [SDK client](https://github.com/slackapi/java-slack-sdk/blob/main/slack-api-model/src/main/java/com/slack/api/model/User.java) entity ```com.slack.api.model.User.Profile```.

It's not a custom field, so as things currently stand I don't think it's possible to set directly via the SDK - it would be great if it could be exposed.

Thanks,
Pete

seratch:
This pull request fixes #857. See the issue for details.

## Requirements

### DIFF
--- a/slack-api-model/src/main/java/com/slack/api/model/User.java
+++ b/slack-api-model/src/main/java/com/slack/api/model/User.java
@@ -74,6 +74,7 @@ public class User {
         private String statusText;
         private String statusTextCanonical;
         private String statusEmoji;
+        private String statusEmojiUrl;
         private Long statusExpiration;
 
         private String displayName;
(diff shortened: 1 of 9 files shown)


slack-api-model/src/main/java/com/slack/api/model/User.java
```diff
@@ -74,6 +74,7 @@
         private String statusText;
         private String statusTextCanonical;
         private String statusEmoji;
+        private String statusEmojiUrl;
```
seratch, line 77: Just moved the field to near the related ones

PR #888, opened 2021-11-11
Fix #887 by skipping to create team_id cache data for metrics

### LINKED ISSUE #887: Using authRevoke() to revoke Slack OIDC token resulting in invalid_auth error
### Reproducible in:

```bash
mvn dependency:tree | grep com.slack.api
gradle dependencies | grep com.slack.api
java -version
sw_vers && uname -v # or `ver`
```

#### The Slack SDK version

```
+--- com.slack.api:bolt-socket-mode:1.13.0
|    +--- com.slack.api:slack-api-model:1.13.0
|    +--- com.slack.api:slack-api-client:1.13.0
|    |    +--- com.slack.api:slack-api-model:1.13.0 (*)
|    +--- com.slack.api:slack-app-backend:1.13.0
|    |    +--- com.slack.api:slack-api-model:1.13.0 (*)
|    |    +--- com.slack.api:slack-api-client:1.13.0 (*)
|    \--- com.slack.api:bolt:1.13.0
|         +--- com.slack.api:slack-api-model:1.13.0 (*)
|         +--- com.slack.api:slack-api-client:1.13.0 (*)
|         \--- com.slack.api:slack-app-backend:1.13.0 (*)
+--- com.slack.api:bolt-socket-mode:1.13.0 (n)
```

#### Java Runtime version

```
openjdk version "11.0.12" 2021-07-20 LTS
OpenJDK Runtime Environment Corretto-11.0.12.7.2 (build 11.0.12+7-LTS)
OpenJDK 64-Bit Server VM Corretto-11.0.12.7.2 (build 11.0.12+7-LTS, mixed mode)
```

#### OS info

```
ProductName:    macOS
ProductVersion: 11.6
BuildVersion:   20G165
Darwin Kernel Version 20.6.0: Mon Aug 30 06:12:21 PDT 2021; root:xnu-7195.141.6~3/RELEASE_X86_64
```

#### Steps to reproduce:

The error occurs when calling `AsyncMethodsClient::authRevoke()` to revoke the OpenID Connect (OIDC) access token. The OIDC access token is granted by the [Sign In With Slack](https://api.slack.com/authentication/sign-in-with-slack) flow.

```java
var slack = new Slack();
var asyncClient = slack.methodsAsync();
asyncClient.authRevoke(req -> req.token("<slack_oidc_access_token>")).join();
```

#### Expected result:

No error occurs. I think this can be fixed either by having the SDK not calling `auth.test` for OIDC access tokens, or for Slack to support `auth.test` for OIDC access tokens.

#### Actual result:

When calling `AsyncMethodsClient::authRevoke()`, I receive the error:

```
Got an unsuccessful response from auth.test API (error: invalid_auth)
at java.lang.Thread.getStackTrace(Thread.java:1598)
    at ch.qos.logback.core.UnsynchronizedAppenderBase.doAppend(UnsynchronizedAppenderBase.java:84)
    at ch.qos.logback.core.spi.AppenderAttachableImpl.appendLoopOnAppenders(AppenderAttachableImpl.java:51)
    at ch.qos.logback.classic.Logger.appendLoopOnAppenders(Logger.java:270)
    at ch.qos.logback.classic.Logger.callAppenders(Logger.java:257)
    at ch.qos.logback.classic.Logger.buildLoggingEventAndAppend(Logger.java:421)
    at ch.qos.logback.classic.Logger.filterAndLog_1(Logger.java:398)
    at ch.qos.logback.classic.Logger.error(Logger.java:526)
    at com.slack.api.methods.impl.TeamIdCache.lambda$lookupOrResolve$0(TeamIdCache.java:44)
    at java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1705)
    at com.slack.api.methods.impl.TeamIdCache.lookupOrResolve(TeamIdCache.java:57)
    at com.slack.api.methods.impl.TeamIdCache.lookupOrResolve(TeamIdCache.java:30)
    at com.slack.api.methods.impl.MethodsClientImpl.postFormWithTokenAndParseResponse(MethodsClientImpl.java:3041)
    at com.slack.api.methods.impl.MethodsClientImpl.authRevoke(MethodsClientImpl.java:1107)
    at com.slack.api.methods.impl.AsyncMethodsClientImpl.lambda$authRevoke$80(AsyncMethodsClientImpl.java:1032)
    at com.slack.api.methods.impl.AsyncRateLimitExecutor.runWithoutQueue(AsyncRateLimitExecutor.java:119)
    at com.slack.api.methods.impl.AsyncRateLimitExecutor.lambda$execute$0(AsyncRateLimitExecutor.java:67)
    at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1771)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
    at java.lang.Thread.run(Thread.java:830)
```

Looking at the code, it seems that the error was logged here:

https://github.com/slackapi/java-slack-sdk/blob/fa458c2f70524b41bd1a6e5014dc3b340f360e3e/slack-api-client/src/main/java/com/slack/api/methods/impl/TeamIdCache.java#L42-L45

`logger.error()` can trigger alerts in our system, so ideally we'd like to have this fixed. I tested sending a request to `https://slack.com/api/auth.test` and it indeeds return `invalid_auth`, which indicates that `auth.test` doesn't support OIDC access token:

```
{
    "ok": false,
    "error": "invalid_auth"
}
```

seratch:
This pull request resolves #887 by updating the internals of `MethodsClient`.

Let me share a bit more about why the `MethodsClient` performs `auth.test` API internally. The Java SDK's API clients support a smart rate limiter out-of-the-box. Refer to the document for more details: https://slack.dev/java-slack-sdk/guides/web-api-basics#rate-limits

To make this feature work with multiple workspace with a single API client instance, the API clients determine `team_id` (this can be `enterprise_id` if the token is issued by org-wide installation) and use it to look up the dedicated thread pool for the workspace/organization. Of course, performing `auth.test` API every time is not optimal in terms of runtime performance. Thus, the API clients manage `team_id` cache data per token.

This pull request disables the metrics data update for the following API methods:
* auth.revoke
* openid.connect.token
* openid.connect.userInfo

I don't think the change in this PR can change the existing behavior. For `auth.revoke`, as the token will be disabled by this call, you don't need to worry about rate limited errors with it any more. For `openid.*` endpoints, the metrics update and auto rate limiter never worked as the team_id resolution does not succeed ("Sign in with Slack" tokens are not available for auth.test API method). For these reasons, I'm sure that it's fine to have the change here in the next minor release.

## Requirements

### DIFF
--- a/bolt-servlet/src/test/java/samples/OpenIDConnectSample.java
+++ b/bolt-servlet/src/test/java/samples/OpenIDConnectSample.java
@@ -49,6 +49,10 @@ public class OpenIDConnectSample {
                     OpenIDConnectUserInfoResponse userInfo = teamIdWiredClient.openIDConnectUserInfo(r -> r
                             .token(refreshedToken.getAccessToken()));
                     logger.info("userInfo: {}", userInfo);
+
+                    // To revoke this, you can use auth.revoke API method
+                    // Slack.getInstance().methods().authRevoke(r -> r.token(refreshedToken.getAccessToken()));
+
                 } else {
                     OpenIDConnectUserInfoResponse userInfo = client.openIDConnectUserInfo(r -> r
                             .token(token.getAccessToken()));
--- a/slack-api-client/src/main/java/com/slack/api/methods/impl/AsyncRateLimitExecutor.java
+++ b/slack-api-client/src/main/java/com/slack/api/methods/impl/AsyncRateLimitExecutor.java
@@ -13,6 +13,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.*;
 
+import static com.slack.api.methods.impl.TeamIdCache.METHOD_NAMES_TO_SKIP_TEAM_ID_CACHE_RESOLUTION;
+
 @Slf4j
 public class AsyncRateLimitExecutor {
 
@@ -60,7 +62,9 @@ public class AsyncRateLimitExecutor {
             Map<String, String> params,
             AsyncExecutionSupplier<T> methodsSupplier) {
         String token = params.get("token");
-        final String teamId = token != null ? teamIdCache.lookupOrResolve(token) : null;
+        final String teamId = (token != null
+                && METHOD_NAMES_TO_SKIP_TEAM_ID_CACHE_RESOLUTION.contains(methodName)) ?
+                teamIdCache.lookupOrResolve(token) : null;
         final ExecutorService executorService = teamId != null ? ThreadPools.getOrCreate(config, teamId) : ThreadPools.getDefault(config);
         return CompletableFuture.supplyAsync(() -> {
             if (NO_TOKEN_METHOD_NAMES.contains(methodName) || teamId == null) {
--- a/slack-api-client/src/main/java/com/slack/api/methods/impl/AsyncRateLimitExecutor.java
+++ b/slack-api-client/src/main/java/com/slack/api/methods/impl/AsyncRateLimitExecutor.java
@@ -63,7 +63,7 @@ public class AsyncRateLimitExecutor {
             AsyncExecutionSupplier<T> methodsSupplier) {
         String token = params.get("token");
         final String teamId = (token != null
-                && METHOD_NAMES_TO_SKIP_TEAM_ID_CACHE_RESOLUTION.contains(methodName)) ?
+                && !METHOD_NAMES_TO_SKIP_TEAM_ID_CACHE_RESOLUTION.contains(methodName)) ?
                 teamIdCache.lookupOrResolve(token) : null;
         final ExecutorService executorService = teamId != null ? ThreadPools.getOrCreate(config, teamId) : ThreadPools.getDefault(config);
         return CompletableFuture.supplyAsync(() -> {
--- a/slack-api-client/src/main/java/com/slack/api/methods/impl/TeamIdCache.java
+++ b/slack-api-client/src/main/java/com/slack/api/methods/impl/TeamIdCache.java
@@ -9,6 +9,8 @@ import okhttp3.FormBody;
 import okhttp3.Response;
 
 import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.function.Function;
@@ -17,6 +19,14 @@ import static com.slack.api.methods.RequestFormBuilder.toForm;
 
 @Slf4j
 public class TeamIdCache {
+
+    // As the tokens issued for "Sign in with Slack" does not work with auth.test API method,
+    // we skip creating team_id cache for these API methods.
+    public static final List<String> METHOD_NAMES_TO_SKIP_TEAM_ID_CACHE_RESOLUTION = Arrays.asList(
+            "auth.revoke",
+            "openid.connect.token",
+            "openid.connect.userInfo"
+    );
 
     private static final ConcurrentMap<String, String> TOKEN_TO_TEAM_ID = new ConcurrentHashMap<>();
 
--- a/slack-api-client/src/main/java/com/slack/api/methods/impl/MethodsClientImpl.java
+++ b/slack-api-client/src/main/java/com/slack/api/methods/impl/MethodsClientImpl.java
@@ -219,6 +219,7 @@ import java.util.*;
 
 import static com.slack.api.methods.RequestFormBuilder.toForm;
 import static com.slack.api.methods.RequestFormBuilder.toMultipartBody;
+import static com.slack.api.methods.impl.TeamIdCache.METHOD_NAMES_TO_SKIP_TEAM_ID_CACHE_RESOLUTION;
 
 @Slf4j
 public class MethodsClientImpl implements MethodsClient {
@@ -3022,7 +3023,8 @@ public class MethodsClientImpl implements MethodsClient {
             String methodName,
             String token) throws IOException {
         String teamId = this.teamId.orElse(null);
-        if (statsEnabled && teamId == null) {
+        if (statsEnabled && teamId == null
+                && !METHOD_NAMES_TO_SKIP_TEAM_ID_CACHE_RESOLUTION.contains(methodName)) {
             teamId = teamIdCache.lookupOrResolve(token);
         }
         try {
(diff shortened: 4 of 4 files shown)


bolt-servlet/src/test/java/samples/OpenIDConnectSample.java
```diff
@@ -49,6 +49,10 @@ public static void main(String[] args) throws Exception {
                     OpenIDConnectUserInfoResponse userInfo = teamIdWiredClient.openIDConnectUserInfo(r -> r
                             .token(refreshedToken.getAccessToken()));
                     logger.info("userInfo: {}", userInfo);
+
+                    // To revoke this, you can use auth.revoke API method
+                    // Slack.getInstance().methods().authRevoke(r -> r.token(refreshedToken.getAccessToken()));
```
seratch, line 54: I did a manual test using this example app. It's not so easy to write meaningful unit tests for this pattern.

slack-api-client/src/main/java/com/slack/api/methods/impl/AsyncRateLimitExecutor.java
```diff
@@ -60,7 +62,9 @@ public static AsyncRateLimitExecutor getOrCreate(MethodsClientImpl client, Slack
             Map<String, String> params,
             AsyncExecutionSupplier<T> methodsSupplier) {
         String token = params.get("token");
-        final String teamId = token != null ? teamIdCache.lookupOrResolve(token) : null;
+        final String teamId = (token != null
+                && METHOD_NAMES_TO_SKIP_TEAM_ID_CACHE_RESOLUTION.contains(methodName)) ?
```
misscoded, line 66: I see that the other checks against `METHOD_NAMES_TO_SKIP_TEAM_ID_CACHE_RESOLUTION` use the negation `!`, so just wanted to verify that this was intended to be the opposite.

bolt-servlet/src/test/java/samples/OpenIDConnectSample.java
misscoded: I take your word for it! It would be nice to test that `lookupOrResolve` isn't invoked if belonging to the skipped methods, but if that was possible I'd imagine you'd have included it.

misscoded (approved):

slack-api-client/src/main/java/com/slack/api/methods/impl/AsyncRateLimitExecutor.java
seratch, line 66: Good catch! This is a bug
```suggestion
                && !METHOD_NAMES_TO_SKIP_TEAM_ID_CACHE_RESOLUTION.contains(methodName)) ?
```

bolt-servlet/src/test/java/samples/OpenIDConnectSample.java
seratch: Good point! Indeed, it would be possible to write such simple test if we refactor the internals of the `MethodsClientImpl` (meaning, enable to pass `TeamIdCache` in some way -- currently it's private with final keyword). But I would like to avoid changing the internals a lot only for this purpose. When we see the need to write the test next time, it'd be a good timing to consider making the part more unit-testable.

slack-api-client/src/main/java/com/slack/api/methods/impl/AsyncRateLimitExecutor.java
seratch: Honestly I haven't verified the async client and overlooked this bug. Thanks for the great catch!

slack-api-client/src/main/java/com/slack/api/methods/impl/TeamIdCache.java
```diff
@@ -18,6 +20,14 @@
 @Slf4j
 public class TeamIdCache {
 
+    // As the tokens issued for "Sign in with Slack" does not work with auth.test API method,
+    // we skip creating team_id cache for these API methods.
+    public static final List<String> METHOD_NAMES_TO_SKIP_TEAM_ID_CACHE_RESOLUTION = Arrays.asList(
+            "auth.revoke",
```
filmaj, line 26: @seratch regarding your comment in this PR:

>  For auth.revoke, as the token will be disabled by this call, you don't need to worry about rate limited errors with it any more. 

It looks like the auth.revoke API accepts a `test` parameter, like a kind of "dry run" execution that wouldn't actually revoke the token. I am not sure if you considered that, but does that change the consideration around removing tracking call frequency for this API?

Acceptable answers include "no, Fil, go away" 🤪

filmaj (approved):
Looks good! Left a question for my own learning purposes. 🙏

slack-api-client/src/main/java/com/slack/api/methods/impl/TeamIdCache.java
seratch: @filmaj haha, of course, I don't say "go away" 😸 

You've pointed out a great point! Indeed, the ignorance here may allow rate limited errors. However, if a developer repeatedly performs auth.revoke for a specific workspace (as this endpoint is [tier 3](https://api.slack.com/docs/rate-limits#tier_t3), specifically 50+ call per minute for a workspace), there should be many other issues in addition to the rate limited error due to the calls. Perhaps, the developer will need to improve the operation to make it more meaningful in the case.

The only realistic use case that I can think of would be a batch processing job that cleans up the tokens in a database and performs a lot of auth.revoke as necessary. 

Indeed, the code with this change cannot handle the situation but I don't think that it's so common. Also, unfortunately, we don't have any means to distinguish whether a token is for normal OAuth or for "Sing in with Slack" without performing auth.test or whatever. For this reason, I think that simply skipping auth.revoke here should be reasonable enough. If the Slack platform provides more light-weight way to verify if a token is for SIWS or not, we can improve this part then.

filmaj: It certainly is an extreme edge case... completely agree it wouldn't be common. I will now go away 😂  Nice work with the PR, thank you for writing this up!

seratch:
@misscoded Thanks again for catching the bug on the async client code!

PR #969, opened 2022-04-13
Fix #966 Ktor incorrectly decoding event json with unicode symbols

### LINKED ISSUE #966: [bolt-ktor] toBoltRequest will receive invalid event request json from enterprise grid workspaces if it contains unicode symbols
There is a bug in Ktor 1.x.x currently (fixed in 2.0.0 I believe) where the behaviour of `call.receiveText()` can potentially lead to invalid request signature matching on event requests sent by an Enterprise Grid workspace.

For example, a lot of Slack users might use the Google/Outlook calendar apps to sync Slack with their meetings. These plugins both set the user status text to be _**In a meeting • Google/Outlook Calendar**_. When this is sent from a normal workspace, the event contains a unicode code point (`\u2022`) for the dot. When sent from a grid, the dot is sent exactly as is. 

The bolt-ktor function `toBoltRequest` uses `call.receiveText()` to get the request body for the event. But because the request does not come with an explicit charset applied to the `Content-Type` header, Ktor's receiveText function ignores the fact that the data is json and uses default plaintext encoding using ISO_8859-1. This turns the dot into `â¢` which changes the bytes. This then fails the request verifier stage.

The workaround is to enforce UTF-8 encoding on requests with a `Content-Type` of `application/json` unless a specific charset is applied. Here is the [Ktor youtrack issue](https://youtrack.jetbrains.com/issue/KTOR-789). We used the following workaround from there:
```kotlin
/**
 * Receive the request as String.
 * If there is no Content-Type in the HTTP header specified use ISO_8859_1 as default charset, see https://www.w3.org/International/articles/http-charset/index#charset.
 * But use UTF-8 as default charset for application/json, see https://tools.ietf.org/html/rfc4627#section-3
 */
private suspend fun ApplicationCall.receiveTextWithCorrectEncoding(): String {
  fun ContentType.defaultCharset(): Charset = when (this) {
    ContentType.Application.Json -> Charsets.UTF_8
    else -> Charsets.ISO_8859_1
  }
  
  val contentType = request.contentType()
  val suitableCharset = contentType.charset() ?: contentType.defaultCharset()
  return receiveStream().bufferedReader(charset = suitableCharset).readText()
}
```

#### The Slack SDK version

Latest (v1.21.1 at time of writing)

#### Java Runtime version
Java 11 (Azul and Coretto) is what we're using

#### Steps to reproduce:

1. Create a simple Slack app using Ktor < 2.0.0
2. Setup app to use user_change event
3. Install the app in an enterprise grid workspace
4. Update your profile status in the enterprise grid to say _**In a meeting • Google Calendar**_

#### Expected result:
Event should be sent to Slack app without issue and not produce a 401 due to invalid signature

#### Actual result:
Request will fail with an invalid request signature and a 401

tg-dclarke:
This pull fixes #966 relating to the bolt-ktor sub module

## Requirements

### DIFF
--- a/bolt-ktor/src/main/kotlin/com/slack/api/bolt/ktor/package.kt
+++ b/bolt-ktor/src/main/kotlin/com/slack/api/bolt/ktor/package.kt
@@ -12,9 +12,12 @@ import io.ktor.http.*
 import io.ktor.request.*
 import io.ktor.response.*
 import io.ktor.util.*
+import io.ktor.utils.io.charsets.Charset
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
 
 suspend fun toBoltRequest(call: ApplicationCall, parser: SlackRequestParser): Request<*>? {
-    val requestBody = call.receiveText()
+    val requestBody = call.receiveTextWithCorrectEncoding()
     val queryString = QueryStringParser.toMap(call.request.queryString())
     val headers = RequestHeaders(call.request.headers.toMap())
     val rawRequest = SlackRequestParser.HttpRequest.builder()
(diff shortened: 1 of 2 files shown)


filmaj (approved):
Thanks so much for sending this patch, this looks good to me! I will wait for @seratch to approve first before merging.

seratch (approved):

PR #989, opened 2022-05-20
Link getting started instructions for gradle to the gradle app quickstart

filmaj:
Using https://docs.gradle.org/current/samples/sample_building_java_applications.html to get a gradle project created helped me run through the getting started guide from start to finish. Might be helpful for total java noobs (like me 😅 ).


seratch (approved):
Thanks!

PR #997, opened 2022-06-15
Add cursor parameter to reactions.list API method

seratch:
This pull request fixes a long-lived bug in reactions.list API support. While an API caller passes count / page, the paging data can be returned, response_metadata.next_cursor can be returned when passing limit parameter instead. This pull request adds response_metadata to the response data class plus add the missing cursor parameter to the request data class.

## Requirements

### DIFF
--- a/json-logs/samples/api/reactions.list.json
+++ b/json-logs/samples/api/reactions.list.json
@@ -1061,8 +1061,7 @@
         "reply_users_count": 12345,
         "latest_reply": "0000000000.000000",
         "reply_users": [
-          "B00000000",
-          "U00000000"
+          "B00000000"
         ],
         "subscribed": false,
         "subtype": "",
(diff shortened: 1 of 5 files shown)


WilliamBergamin:
Everything looks good on the sdk side of things, the tests I ran earlier are working now :)

My tests did raise a concern about the behavior of the API, it seem like the `cursor` field in the returned json payload is set to an empty string instead of being not present or null.

Here is an example of a completed paging request:
```json
{
    "ok": true,
    "items": [
        ...
    ],
    "response_metadata": {
        "next_cursor": ""
    }
}
```

This leads to this project also setting the cursor to an empty String, I don't know if we want to add a special case for this.

seratch:
@WilliamBergamin 
>My tests did raise a concern about the behavior of the API, it seem like the cursor field in the returned json payload is set to an empty string instead of being not present or null.

This is an expected behavior if your workspace/user does not have any further responses. When it comes to the robustness of the integration tests (meaning making them to run with any workspaces etc.), we can improve it later once you started actively working on this project.

seratch:
Thanks for your review 👍 Let me merge this PR now.

PR #1004, opened 2022-06-21
Arraylist<LayoutBlock> GSON anonymous inner class initialization bug

WilliamBergamin:
GSON has a [known issue](https://github.com/google/gson/issues/2023#issuecomment-974230703) where null is returned if it tries to serialize an object that was initialized using an anonymous inner class.

Example:
```java
import com.google.gson.Gson;
import com.google.gson.GsonBuilder;

import java.util.ArrayList;
import java.util.List;

List<String> strings = new ArrayList<>() {{
            add("hello");
            add("world");
        }};

GsonBuilder builder = new GsonBuilder();
Gson gson = builder.create();

String json = gson.toJson(strings);

assertEquals("{\"hello\", \"world\"}", json); // this should be true but json = "null"
```

The above assertion should be true but the json string generated by the GSON lib actually returns a string of value "null". This can lead to some head scratching for the users of this library.

These changes try to reinitialize & reserialize the list of blocks if this "null" output is obtained from the GSON library.

## Requirements

### DIFF
--- a/slack-api-client/src/main/java/com/slack/api/methods/RequestFormBuilder.java
+++ b/slack-api-client/src/main/java/com/slack/api/methods/RequestFormBuilder.java
@@ -2496,6 +2496,8 @@ public class RequestFormBuilder {
     private static final String FALLBACK_WARN_MESSAGE_TEMPLATE =
             "Additionally, the attachment-level `fallback` argument is missing in the request payload for a {} call - To avoid this warning, it is recommended to always provide a top-level `text` argument when posting a message. Alternatively, you can provide an attachment-level `fallback` argument, though this is now considered a legacy field (see https://api.slack.com/reference/messaging/attachments#legacy_fields for more details).";
 
+    private static final String GSON_ANONYM_INNER_CLASS_INIT_OUTPUT = "null";
+
     private static void warnIfAttachmentWithoutFallbackDetected(String endpointName, List<Attachment> attachments) {
         boolean fallbackMissing = false;
         for (Attachment a : attachments) {
@@ -2556,7 +2558,7 @@ public class RequestFormBuilder {
 
     private static String getJsonWithGsonAnonymInnerClassHandling(List<LayoutBlock> blocks){
         String json = GSON.toJson(blocks);
-        if ("null".equals(json)){
+        if (GSON_ANONYM_INNER_CLASS_INIT_OUTPUT.equals(json)){
             return GSON.toJson(new ArrayList<>(blocks));
         }
         return json;
--- a/slack-api-client/src/main/java/com/slack/api/methods/RequestFormBuilder.java
+++ b/slack-api-client/src/main/java/com/slack/api/methods/RequestFormBuilder.java
@@ -108,6 +108,7 @@ import com.slack.api.methods.request.workflows.WorkflowsStepFailedRequest;
 import com.slack.api.methods.request.workflows.WorkflowsUpdateStepRequest;
 import com.slack.api.model.Attachment;
 import com.slack.api.model.ConversationType;
+import com.slack.api.model.block.LayoutBlock;
 import com.slack.api.util.json.GsonFactory;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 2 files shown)


slack-api-client/src/main/java/com/slack/api/methods/RequestFormBuilder.java
```diff
@@ -2553,4 +2556,12 @@ private static void setIfNotNull(String name, Object value, MultipartBody.Builde
         }
     }
 
+    private static String getJsonWithGsonAnonymInnerClassHandling(List<LayoutBlock> blocks){
```
filmaj, line 2559: I'd add a comment here that explains _why_ we need this method, just as a hint to a maintainer reading through this code in the future (I have a bad memory so I probably won't remember the details of this problem in 6 months!). Maybe just a quick link to the GSON issue and a single sentence explanation.

filmaj (approved):
LGTM, but will defer to @seratch for approval

slack-api-client/src/test/java/test_locally/api/methods/RequestFormBuilderTest.java
```diff
@@ -0,0 +1,81 @@
...
+    @AllArgsConstructor
+    private static class MyTestBlock implements LayoutBlock {
+        public static final String TYPE = "myTestBlock";
+        private final String type = TYPE;
+        private String blockId;
+    }
+
+    @Test
+    public void testGetJsonWithGsonAnonymInnerClassHandling() throws NoSuchMethodException, InvocationTargetException, IllegalAccessException {
+        // SETUP
```
seratch, line 34: I like the test structure. If you want to have these comments, probably Given-When-Then style would be more common.

slack-api-client/src/main/java/com/slack/api/methods/RequestFormBuilder.java
seratch: Sounds good. Adding a link to the gson project issue would be easier.

slack-api-client/src/test/java/test_locally/api/methods/RequestFormBuilderTest.java
```diff
@@ -0,0 +1,81 @@
...
+        Method method = RequestFormBuilder.class.getDeclaredMethod("getJsonWithGsonAnonymInnerClassHandling", List.class);
```
seratch, line 35: You can test one of the public methods that internally calls getJsonWithGsonAnonymInnerClassHandling instead of trying to run the private method using reflection APIs.

slack-api-client/src/main/java/com/slack/api/methods/RequestFormBuilder.java
```diff
@@ -1134,7 +1135,7 @@ public static FormBody.Builder toForm(ChatScheduleMessageRequest req) {
         if (req.getBlocksAsString() != null) {
             form.add("blocks", req.getBlocksAsString());
         } else if (req.getBlocks() != null) {
-            String json = GSON.toJson(req.getBlocks());
+            String json = getJsonWithGsonAnonymInnerClassHandling(req.getBlocks());
```
seratch, line 1138: `req.getUsers(): List<String>` patterns also need the same workaround.

seratch (changes requested):

slack-api-client/src/main/java/com/slack/api/methods/RequestFormBuilder.java
WilliamBergamin: Agreed, I added a comment with the link to the original issue

slack-api-client/src/main/java/com/slack/api/methods/RequestFormBuilder.java
WilliamBergamin: Makes sense, I also noticed that the Map objects also suffered from this GSON issue, I added additional code to handle these use cases.

seratch (approved):
Great improvement 💯  Thanks for fixing this!

PR #1012, opened 2022-07-08
Updates `tyrus-standalone-client` version from 1.17 to 1.19 in the documentations

marcwrobel:
This PR updates `tyrus-standalone-client` version from 1.17 to 1.19 in the documentations. `tyrus-standalone-client` 1.19 is still compatible with bolt-socket-mode, but supports officially Java 17 and contains various bug fixes and improvements. See https://projects.eclipse.org/projects/ee4j.tyrus/tyrus-standalone-client.

The PR has been split in two commits so that even if the `tyrus-standalone-client` version update is not possible future version updates will be easier for both  `javax.websocket-api` and `tyrus-standalone-client` :
1. Replace `javax.websocket-api` and `tyrus-standalone-client` versions by variables in documentation. Note that a few trailing whitespaces were automatically trimmed and a few final new lines were automatically added. This is not an issue so it has not been reverted.
2. Update `tyrus-standalone-client` version in documentation.

## Requirements

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -15,4 +15,4 @@ compatibleMicronautVersion: 3.x
 quarkusVersion: 2.10.2.Final
 helidonVersion: 2.5.0
 javaxWebsocketApiVersion: 1.1
-tyrusStandaloneClientVersion: 1.17
+tyrusStandaloneClientVersion: 1.19
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -14,3 +14,5 @@ springBootVersion: 2.7.1
 compatibleMicronautVersion: 3.x
 quarkusVersion: 2.10.2.Final
 helidonVersion: 2.5.0
+javaxWebsocketApiVersion: 1.1
+tyrusStandaloneClientVersion: 1.17
--- a/docs/guides/getting-started-with-bolt-socket-mode.md
+++ b/docs/guides/getting-started-with-bolt-socket-mode.md
@@ -46,12 +46,12 @@ After you [create your Maven project](https://maven.apache.org/guides/getting-st
 <dependency>
   <groupId>javax.websocket</groupId>
   <artifactId>javax.websocket-api</artifactId>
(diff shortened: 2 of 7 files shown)


WilliamBergamin:
Hi @marcwrobel thanks for your contribution & thank you for spotting this issue in our docs!

I tested this on my local and your variable reference changes work as intended 👍 

That being said, I am uncertain that this is a behavior we want to adopt. As writing our documentation in markdown and rendering it with  jekyll  serves a dual purpose:
1. The documentation can be viewed [through our website](https://slack.dev/java-slack-sdk/guides/socket-mode)
2. The documentation can be viewed through the [github repo via vanilla markdown](https://github.com/slackapi/java-slack-sdk/blob/v1.23.0/docs/guides/socket-mode.md#project-setup)

When performing this reference operation we lose readability in the github repo since the `version value` is not rendered properly (aka the variable name is displayed in vanilla markdown)

I tend to value both these functionalities and would be in favor of not using `variable references` in the doc. This way the markdown renders remain simple and clear. Maybe @seratch could add additional input on this.

marcwrobel:
Hi @WilliamBergamin, that is just my opinion but I would agree with you if those versions was referenced only twice (once for each translation). But in the doc both version are referenced twelve times (six time for each translation).

So the question at hand is : should you favor maintenance and consistency, or readability ?

seratch:
@WilliamBergamin 
Indeed, some people may want to read the vanilla markdown but I don't think it is a common need. Also, since we already have some others [here](https://github.com/slackapi/java-slack-sdk/blob/main/docs/_config.yml), there is no reason to avoid using variables only for these repetitive references for Socket Mode dependencies.

For above reasons, my opinion is that we can (and should) merge this improvement by @marcwrobel 
@WilliamBergamin Are you fine with this?

seratch (approved):
LGTM

WilliamBergamin (approved):
LGTM

Issue #1055, opened 2022-09-22
Is there a way how to list all enterprise and team ID of the installed applications?


musketyr:
I'm working on example with a distributed application. I know I can use `InstallationService` to get the bot information but is there
a method to list all the installed bots somewhere?

If not, what is the way how to hook into the installation process to store the bot information separately?

#### The Slack SDK version

`1.25.1`

#### Java Runtime version

```
openjdk version "11.0.16" 2022-07-19 LTS
OpenJDK Runtime Environment Zulu11.58+15-CA (build 11.0.16+8-LTS)
OpenJDK 64-Bit Server VM Zulu11.58+15-CA (build 11.0.16+8-LTS, mixed mode)
```

#### OS info

```
ProductName:    macOS
ProductVersion: 12.2.1
BuildVersion:   21D62
Darwin Kernel Version 21.3.0: Wed Jan  5 21:37:58 PST 2022; root:xnu-8019.80.24~20/RELEASE_ARM64_T6000
```

#### Steps to reproduce:

1. Create a distributed app
2. Install the app into the workspace

#### Expected result:

There is a method which can list all the available installed bots

#### Actual result:

I'm unable to locate such method.

### Requirements

Please make sure if this topic is specific to this SDK. For general questions/issues about Slack API platform or its server-side, could you submit questions at https://my.slack.com/help/requests/new instead. :bow:

Please read the [Contributing guidelines](https://github.com/slackapi/java-slack-sdk/blob/main/.github/contributing.md) and [Code of Conduct](https://slackhq.github.io/code-of-conduct) before creating this issue or pull request. By submitting, you agree to those rules.

hello-ashleyintech:
Hi, @musketyr! Thank you for submitting your question! 🙌 

If you are only looking for installed bots and not installed applications, you can use the [`users.list`](https://api.slack.com/methods/users.list) API method to achieve what you're looking for. Since all bots have bot users, they will show up in this list of all users in your workspace. You'll need to iterate through the returned list of users and also check for the `is_bot` property for each entry in the response.

musketyr:
Hello @hello-ashleyintech,

[...]

PR #1056, opened 2022-09-23
Find all installed bots

musketyr:
This PR adds a method into `InstallationService` to list all the installed bots . See #1055 for further Information.

## Requirements


seratch:
Hi @musketyr, thanks for sending this PR! However, we are not planning to add this method to the interface. Please consider having your own sub-class of `AmazonS3InstallationService` with the additional method plus using its instance in your listener methods.

musketyr:
if you agree on this addition, I can add test using [Testcontainers with Localstack](https://www.testcontainers.org/modules/localstack/) for the Amazon S3 implementation.

musketyr:
@seratch ok, this is quite a pity because it would make more sense to have the the ability directly in the library.

seratch:
@musketyr Unfortunately, I don't think that this is a common scenario in many Slack apps while the find-all method can be useful for you. I'd appreciate it if you could understand our stance on maintaining this SDK here.

PR #1064, opened 2022-09-29
Fix #1060 Enable Slack#close() method to shutdown thread pools behind its SlackConfig

### LINKED ISSUE #1060: Enable Slack#close() method to shutdown thread pools behind its SlackConfig
Hello,

I am not sure whether this is a bug or really some misuse of the Slack SDK by my side.

I stumbled into an issue where a long-running JakartaEE-based application was getting overwhelmed with slack-api-metrics threads.
By the two week mark it had over 3900 of these threads, despite all of them being parked.

After some digging through the code and some debugging I realised that these threads are created everytime I instantiate a SlackConfig object which I did everytime I needed to call Slack API throught the SDK to send a message to a conversation.

Although I wasn't aware that these threads were created at all, I was expecting that by closing the Slack instance, these kind of resources could be freed or at least that the SDK realised that there were already some threads being used.

Below is the general aspect of the code I was using. Some time during the normal operation of the application there would be a need to send some messages, so I would invoke an EJB that would create a SlackConfig object (with slightly different settings), would get a Slack instance with that config and send the messages, closing it afterwards.

### Reproducible in:

```
private void sendMessagesToChannel(List<SlackMessageDTO> slackMessages) {

        SlackConfig slackConfig = new SlackConfig();
        slackConfig.setHttpClientWriteTimeoutMillis(SECURITY_MARGIN_IN_MILLISSECONDS);
        slackConfig.setHttpClientReadTimeoutMillis(SECURITY_MARGIN_IN_MILLISSECONDS);
        slackConfig.setHttpClientCallTimeoutMillis(slackMessages.size() * SECURITY_MARGIN_IN_MILLISSECONDS);

        try (Slack slack = Slack.getInstance(slackConfig)) { 
            for (SlackMessageDTO slackMessage : slackMessages) {

                // Compose message

               // Send message
                slack.methods(slackToken)
                    .chatPostMessage(req -> req
                        .channel(slackMessage.channel)
                        .attachments(Collections.singletonList(attachment))
                        .text(slackMessage.message));
            }
        } catch (Exception e) {
           // Log issue
        }

    }

```

What happens here is:

1. the SDK creates an Executor with some initial worker (I think 6-9) threads called **slack-api-metrics**
2. everytime new messages are sent through that code, on each new SlackConfig object instantiated a further 3 new **slack-api-metrics** threads are created and added to the Executor
3. never terminates any of those, even when the close(). is called on Slack object, because that frees only the underlying OkHttpClient resources.

Naturally, when I realised that 3 slack-api-metrics threads were being created everytime I created a SlackConfig object (one thread for MethodsConfig, AuditConfig and SCIMConfig) I solved the problem by just having a single SlackConfig object for the whole application that would be used whenever I needed a new Slack instance.

However, I am wondering if:

- it was only ever meant to be used one instance of SlackConfig on the entire application (if so, I think it would be useful to point that out somewhere, maybe in Javadoc) and maybe enforce it as a Singleton
- if one can create multiple instances of SlackConfig then somehow it should clean its resources after the associated instance was closed, otherwise it can happen this kind of thread pollution in long-running apps
- if I disable these metrics (one can do since recently) what will I lose? Only metric collection? Or does that impact other kind of functionality?
- Also, one cannot completetely disable these metrics as stated here https://github.com/slackapi/java-slack-sdk/issues/987

Thanks in advance.

#### The Slack SDK version

**1.25.1**

#### Java Runtime version

OpenJDK 64-Bit Server VM Corretto-11.0.16.9.1

#### OS info

Windows 10 Enterprise

#### Steps to reproduce:

Well, the kind of code I posted there should do it. Sorry for not providing a reproducer project.

#### Expected result:

Created metrics threads should be removed after the Slack instance is closed.

#### Actual result:

The JVM keeps the created threads and adds 3 more threads per each call

### Requirements
 N/A

seratch:
This pull request resolves #1060 by enhancing Slack and SlackConfig classes to shutdown thread pools once close() method is called. Also, I've updated the javadoc of the SlackConfig class to clearly mention that reusing the instance is recommended.

To reviewers:

Please don't feel pressured even if you are not familiar with the internals of this SDK. To learn how SlackConfig and its sub config classes are connected to the ThreadPools classes, you can check how `ThreadPools.getOrCreate()` method works.

## Requirements

### DIFF
--- a/slack-api-client/src/main/java/com/slack/api/SlackConfig.java
+++ b/slack-api-client/src/main/java/com/slack/api/SlackConfig.java
@@ -23,9 +23,21 @@ import java.util.Map;
 
 /**
  * The basic configuration of this SDK. Some settings can be propagated to submodules such as Bolt.
+ * <p>
+ * Please note that, if you are fine with the same settings for all Slack instances across an application,
+ * using a singleton instance of this config class is highly recommended.
+ * Also, a Slack instance is thread-safe, so you can use singleton for it too.
+ * <p>
+ * If you create a new SlackConfig instance for each Slack instance creation, each SlackConfig object
+ * can create thread pools for maintaining metrics data and async executions.
+ * In most use cases, this should not be intended. To avoid this, consider reusing SlackConfig objects
+ * as much as possible.
+ * <p>
+ * An alternative way is to set the statsEnabled flag to false.
+ * As long as you don't use the metrics and async API calls at all, there is no downside by doing so.
  */
 @Data
-public class SlackConfig {
+public class SlackConfig implements AutoCloseable {
 
     /**
      * The default instance is immutable. It's not allowed to modify the value runtime for any reasons.
@@ -34,6 +46,11 @@ public class SlackConfig {
 
         void throwException() {
             throw new UnsupportedOperationException("This config is immutable");
+        }
+
+        @Override
+        public void close() {
+            // We disable closing resources for this default singleton object to avoid misbehaviors
         }
 
         @Override
@@ -342,4 +359,10 @@ public class SlackConfig {
         sCIMConfig.getMetricsDatastore().setTraceMode(this.isLibraryMaintainerMode());
     }
 
+    @Override
+    public void close() throws Exception {
+        com.slack.api.methods.impl.ThreadPools.shutdownAll(methodsConfig);
+        com.slack.api.audit.impl.ThreadPools.shutdownAll(auditConfig);
+        com.slack.api.scim.impl.ThreadPools.shutdownAll(sCIMConfig);
+    }
 }
--- a/slack-api-client/src/main/java/com/slack/api/Slack.java
+++ b/slack-api-client/src/main/java/com/slack/api/Slack.java
@@ -113,7 +113,12 @@ public class Slack implements AutoCloseable {
 
     @Override
     public void close() throws Exception {
-        getHttpClient().close();
+        if (getHttpClient() != null) {
+            getHttpClient().close();
+        }
+        if (getConfig() != null) {
+            getConfig().close();
+        }
     }
 
     /**
--- a/slack-api-client/src/main/java/com/slack/api/audit/impl/ThreadPools.java
+++ b/slack-api-client/src/main/java/com/slack/api/audit/impl/ThreadPools.java
@@ -36,6 +36,25 @@ public class ThreadPools {
         }
     }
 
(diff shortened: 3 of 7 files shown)


seratch:
The CI builds seem to be detecting some race condition issues:
```
closeMethod(test_locally.SlackConfigTest)  Time elapsed: 0.017 sec  <<< FAILURE!
java.lang.AssertionError: 
Expected: is <true>
     but: was <false>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:8)
	at test_locally.SlackConfigTest.closeMethod(SlackConfigTest.java:167)
```

seratch:
^ the above issue was due to the test code quality. improved.

WilliamBergamin (approved):
Nice use of `AutoCloseable`, this looks good to me 💯

slack-api-client/src/main/java/com/slack/api/SlackConfig.java
```diff
@@ -23,9 +23,21 @@
...
+ * Please note that, if you are fine with the same settings for all Slack instances across an application,
+ * using a singleton instance of this config class is highly recommended.
+ * Also, a Slack instance is thread-safe, so you can use singleton for it too.
+ * <p>
+ * If you create a new SlackConfig instance for each Slack instance creation, each SlackConfig object
+ * can create thread pools for maintaining metrics data and async executions.
+ * In most use cases, this should not be intended. To avoid this, consider reusing SlackConfig objects
+ * as much as possible.
+ * <p>
+ * An alternative way is to set the statsEnabled flag to false.
```
filmaj, line 36: I just realized that these docs are hosted on sonatype.org, and as far as I can tell, can't be previewed locally in Jekyll 😢 .
I wanted to check to see if the references to various classes (i.e. `SlackConfig`) and maybe even properties (i.e. `statsEnabled`) would link properly. I suppose we won't be able to tell until we merge + deploy?

filmaj (approved):
LGTM! I today learned about `AutoCloseable` 🙇

slack-api-client/src/main/java/com/slack/api/SlackConfig.java
seratch: FYI: You can package javadoc.jar files by running mvn command and can unzip the file to check the generated HTML!

PR #1076, opened 2022-10-27
Add datetimepicker, url, email, number block elements

seratch:
This pull request adds four block elements, which were added in October 2022.

* datetimepicker
* url_text_input
* email_text_input
* number_input

## Requirements


slack-api-model/src/main/java/com/slack/api/model/block/element/NumberInputElement.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * The initial value in the plain-text input when it is loaded.
+     */
+    private String initialValue;
+
+    /**
+     * The minimum value, cannot be greater than max_value.
+     */
+    private String minValue;
```
WilliamBergamin, line 48: In the docs this is a `String` but should this potentially be a `float`?

WilliamBergamin (approved):
This looks good to me 🚀  just had one question

slack-api-model/src/main/java/com/slack/api/model/block/element/NumberInputElement.java
filmaj: I brought this up internally in Slack. See the thread on this topic in our team channel: https://slack-pde.slack.com/archives/GD7351R97/p1666888618115249

seratch: @WilliamBergamin @filmaj I agree that it can be float and we can decide after our internal discussion. One thing I wanted to mention is that, even if we decide to change this side, the view_submission's state.values still return a string value and perhaps it may be too late to change the view_submission payload side. If so, for consistency, I think that going with String for these is not that bad while it's not developer-friendly.

@filmaj I know that just sharing the internal Slack permalink itself is not that risky but I have a bit strong opinion that we should not expose any internal URLs in public places because such a behavior can be viewed by external people as lack of consideration of security. I prefer how we can be viewed than our own convenience. You may have a different opinion here (and it's fine!). If so, let's discuss it separately.

slack-api-model-kotlin-extension/src/main/kotlin/com/slack/api/model/kotlin_extension/block/element/NumberInputElementBuilder.kt
```diff
@@ -0,0 +1,112 @@
...
+     * Only one element can be set to true. Defaults to false.
+     *
+     * @see <a href="https://api.slack.com/reference/block-kit/composition-objects#dispatch_action_config">The document</a>
+     */
+    fun focusOnLoad(focusOnLoad: Boolean) {
+        _focusOnLoad = focusOnLoad
+    }
+
+    /**
+     * Indicates whether the element will be set to autofocus within the view object.
```
filmaj, line 90: I think this is the wrong document for this method.

slack-api-model/src/main/java/com/slack/api/model/block/element/NumberInputElement.java
filmaj: True, there are some internal backend implementation considerations the API team has to balance and thus this is how we ended up with a string type for these properties. @WilliamBergamin and I discussed with them yesterday about the path forward for the HTTP API. It seems for the time being the HTTP API will remain as is.

If we want, we _could_ implement different types in our SDKs to match the "expectation" that developers may have around these properties, if we wish. E.g. in the Java implementation maybe we could overload the setter methods for the various number input `*_value` properties to accept either integer or float, and then implement logic to map to a string as part of serialization. This would effectively hide these less optimal parts of the API from the developer. However, this then starts to create a separation between the underlying HTTP API and the SDKs, and the need to maintain these exceptional code paths. Possibly it is not worth it. I am open to either approach, just raising it as an option, wondering what others think.

@seratch my apologies, good to know re: sharing URLs. I have edited my comment and will not do so in the future.

seratch: @filmaj 
Interesting. This type of overloading can be very complex in Java as the lang has static types plus does not have union types. It may be possible to have value and valueAsInt, valueAsFloat as a workaround but I don't think this is intuitive and perhaps many devs won't use it. Also, the JSON serialization process can be a bit complicated. For these reasons, regardless of the outcomes of our discussion for other SDKs, I am down to supporting only string for Java SDK.

filmaj: Makes sense to me @seratch, thanks for explaining!

PR #1081, opened 2022-11-08
Add requestFileInfo param to MethodsClient#filesUploadV2 method

seratch:
This pull request adds requestFileInfo option to MethodsClient#filesUploadV2 method for feature parity with Node and Python SDKs. See the following PRs for its context:
* https://github.com/slackapi/python-slack-sdk/pull/1282
* https://github.com/slackapi/node-slack-sdk/pull/1544

## Requirements

### DIFF
--- a/slack-api-client/src/main/java/com/slack/api/methods/impl/FilesUploadV2Helper.java
+++ b/slack-api-client/src/main/java/com/slack/api/methods/impl/FilesUploadV2Helper.java
@@ -10,6 +10,7 @@ import com.slack.api.methods.response.files.FilesCompleteUploadExternalResponse;
 import com.slack.api.methods.response.files.FilesGetUploadURLExternalResponse;
 import com.slack.api.methods.response.files.FilesInfoResponse;
 import com.slack.api.methods.response.files.FilesUploadV2Response;
+import com.slack.api.model.File;
 import com.slack.api.util.http.SlackHttpClient;
 import lombok.Data;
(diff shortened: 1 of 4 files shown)


seratch:
test

slack-api-client/src/main/java/com/slack/api/methods/impl/FilesUploadV2Helper.java
```diff
@@ -106,12 +107,17 @@ public FilesUploadV2Response completeUploads(
 
         result.setFiles(new ArrayList<>());
         for (FilesCompleteUploadExternalResponse.FileDetails file : response.getFiles()) {
-            FilesInfoResponse fileInfo = this.client.filesInfo(r -> r.token(v2Request.getToken()).file(file.getId()));
-            underlyingException.getFileInfoResponses().add(fileInfo);
-            if (!fileInfo.isOk()) {
-                throw underlyingException;
+            if (v2Request.isRequestFileInfo()) {
```
WilliamBergamin, line 110: NIT: I would tend to try and place this in its own function named something like `fetchFile` that returns the `File` object from `fileInfo.getFile()` or `partialFileObject`

seratch: Thanks for the suggestion. In this case, the part won't be reused elsewhere plus we have to pass v2Request 
 and underlyingException to the internal method, which is not so simple. So, let me go with the current code.

PR #1086, opened 2022-11-16
Fix #1084 by enabling app.message method to manage multiple handlers

### LINKED ISSUE #1084: Calling App.message() will overwrite the previous App.message()
example:

```java
App app = new App();
app.message("pattern1", (event, ctx) -> {
    // block1
});
app.message("pattern2", (event, ctx) -> {
    // block2
});
```
As shown in the above code, "pattern1" will be overwritten by "pattern2" and become invalid.

seratch:
This pull request resolves #1084 by enabling `app.message` method to internally manage multiple handler functions. With that, `App` can apply multiple registered message event handlers one by one until it finds a matching pattern.

## Requirements

### DIFF
--- a/bolt/src/main/java/com/slack/api/bolt/App.java
+++ b/bolt/src/main/java/com/slack/api/bolt/App.java
@@ -176,8 +176,10 @@ public class App {
 
     /**
      * Registered Event API handlers.
+     * Note that only limited set of events allow having multiple handlers.
+     * As of v1.27.1, only MessageEvent.class ones can be multiple.
      */
-    private final Map<String, BoltEventHandler<Event>> eventHandlers = new HashMap<>();
+    private final Map<String, List<BoltEventHandler<Event>>> eventHandlers = new HashMap<>();
 
     /**
      * Primitive Event API handler.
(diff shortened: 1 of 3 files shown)


bolt/src/main/java/com/slack/api/bolt/App.java
```diff
@@ -634,15 +653,15 @@ public App message(String pattern, BoltEventHandler<MessageEvent> messageHandler
...
+        event(MessageEvent.class, true, (event, ctx) -> {
             String text = event.getEvent().getText();
             if (log.isDebugEnabled()) {
                 log.debug("Run a message event handler (pattern: {}, text: {})", pattern, text);
             }
             if (text != null && pattern.matcher(text).matches()) {
                 return messageHandler.apply(event, ctx);
             } else {
-                return ctx.ack();
+                return null;
```
seratch, line 664: This is a behavior change for unmatched patterns but I would call it a bug fix rather than a breaking change. If the message event handler does not match, bolt-java should not acknowledge the request. This is a major difference from bolt-js in its design. If we receive more feedback on the consistent behavior with bolt-js in the future, it's relatively easy to introduce a new option to turn on "auto-acknowledgement" for all Events API requests like bolt-js does. However, this is out of this PR's scope.

bolt/src/main/java/com/slack/api/bolt/App.java
```diff
@@ -1346,10 +1370,15 @@ protected Response runHandler(Request slackRequest) throws IOException, SlackApi
                         return Response.ok();
                     }
                     EventRequest request = new EventRequest(stepRequest.getRequestBodyAsString(), stepRequest.getHeaders());
-                    BoltEventHandler<Event> handler = eventHandlers.get(request.getEventTypeAndSubtype());
-                    if (handler != null) {
-                        EventsApiPayload<Event> payload = buildEventPayload(request);
-                        return handler.apply(payload, request.getContext());
+                    List<BoltEventHandler<Event>> handlers = eventHandlers.get(request.getEventTypeAndSubtype());
```
seratch, line 1373: This duplicated code is for Steps from apps (the legacy workflow enhancement). The code can be improved not to repeat the same lines of code but not this time.

bolt/src/test/java/test_locally/app/MessageTest.java
```diff
@@ -80,7 +83,7 @@ public void withPayload_skipped() throws Exception {
 
         EventRequest req = buildRequest(messagePayload);
         Response response = app.run(req);
-        assertEquals(200L, response.getStatusCode().longValue());
+        assertEquals(404L, response.getStatusCode().longValue());
```
seratch, line 86: As mentioned above, this is an intentional behavior change

PR #1181, opened 2023-07-07
Fix #1179 Add message_blocks property to Attachement class

### LINKED ISSUE #1179: Add message_blocks property to Attachement class
Hi to everyone!
I am trying to use bolt java library to integrate forward information in my system. I have some problems to access the right field in the object. First of all I fetch the message using Conversations.history (https://api.slack.com/methods/conversations.history)
Then if the message has a forward in the json response there is a "attachments" field that contains a list of attachment. Now in the tester the json response has a "message_blocks" field, that is basically what I need because I need to parse one by one the blocks in the message object contained in the message_blocks field. The problem is that the "message_blocks field is not present in the bolt Attachment class
https://javadoc.io/doc/com.slack.api/slack-api-model/1.1.1/com/slack/api/model/Attachment.html
https://github.com/slackapi/java-slack-sdk/blob/main/slack-api-model/src/main/java/com/slack/api/model/Attachment.java
 
It seems like that this field is not mapped
 
I would really like to access that field or find another way to get the blocks information about the forwarded message.
 
Thank you in advance

seratch:
This pull request resolves #1179 

## Requirements


zubchenok:
this change causes exception. message_blocks of type LayoutBlock cannot be read, because there is no type field in message_blocks.

PR #1268, opened 2024-01-22
Correct the rate limit tier setting for `users.profile.set` API calls

gunrein:
Thanks for providing this client library!

We have recently had some trouble with rate limiting on the `users.profile.set` method. This PR makes two changes.

1. It fixes an apparent bug in the async `methods` client where the queue depth for a method is reset to 0 when `AsyncRateLimitExecutor.execute` is called. There is a separate thread that will update the queue depth metric, but ideally it is never reset to 0 incorrectly. When the metric is reset it can lead to no (or incorrect) backoff being calculated in `WaitTimeCalculator.calculateWaitTime`.
2. It changes the rate limit tier for `users.profile.set` to more closely match the documentation. https://api.slack.com/apis/rate-limits#profile_updates indicates the rate limit is 30 requests per minute so `Tier2` seems more appropriate / safer.

## Requirements

### DIFF
--- a/slack-api-client/src/main/java/com/slack/api/methods/MethodsRateLimits.java
+++ b/slack-api-client/src/main/java/com/slack/api/methods/MethodsRateLimits.java
@@ -435,7 +435,7 @@ public class MethodsRateLimits {
         setRateLimitTier(USERS_SET_PHOTO, Tier2);
         setRateLimitTier(USERS_SET_PRESENCE, Tier2);
         setRateLimitTier(USERS_PROFILE_GET, Tier4);
-        setRateLimitTier(USERS_PROFILE_SET, Tier4);
+        setRateLimitTier(USERS_PROFILE_SET, Tier2);
 
         setRateLimitTier(VIEWS_OPEN, Tier4);
         setRateLimitTier(VIEWS_PUBLISH, Tier4);
--- a/slack-api-client/src/main/java/com/slack/api/methods/MethodsRateLimits.java
+++ b/slack-api-client/src/main/java/com/slack/api/methods/MethodsRateLimits.java
@@ -435,7 +435,7 @@ public class MethodsRateLimits {
         setRateLimitTier(USERS_SET_PHOTO, Tier2);
         setRateLimitTier(USERS_SET_PRESENCE, Tier2);
         setRateLimitTier(USERS_PROFILE_GET, Tier4);
-        setRateLimitTier(USERS_PROFILE_SET, Tier2);
+        setRateLimitTier(USERS_PROFILE_SET, Tier3);
 
         setRateLimitTier(VIEWS_OPEN, Tier4);
         setRateLimitTier(VIEWS_PUBLISH, Tier4);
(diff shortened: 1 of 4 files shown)


seratch:
Hi @gunrein, thank you so much for taking the time to send this PR. It appears that the rate limit for users.profile.set may have loosened from Tier 4 to Tier 2 at some point, therefore the outdated configuration in the code should be corrected. However, your change to the internal method for calculating queue size isn't necessary at this time. Would you mind reverting that change? Indeed, the method signature could be somewhat confusing, but it does work without any issues.

slack-api-client/src/main/java/com/slack/api/methods/MethodsRateLimits.java
```diff
@@ -435,7 +435,7 @@ public static void setRateLimitTier(String methodName, MethodsRateLimitTier tier
         setRateLimitTier(USERS_SET_PHOTO, Tier2);
         setRateLimitTier(USERS_SET_PRESENCE, Tier2);
         setRateLimitTier(USERS_PROFILE_GET, Tier4);
-        setRateLimitTier(USERS_PROFILE_SET, Tier4);
+        setRateLimitTier(USERS_PROFILE_SET, Tier2);
```
seratch, line 438: The difference between documents could be confusing but the rate limit is Tier 3 on the API endpoint side: https://api.slack.com/methods/users.profile.set Thus, we'd like to use the setting (Tier 3) for it.

seratch (changes requested):

slack-api-client/src/main/java/com/slack/api/methods/MethodsRateLimits.java
gunrein: Sounds good, will do.

gunrein:
Great, thanks @seratch.

I'll back out the change to the internal method. One question on that, though (feel free to tell me it isn't worth explaining; no worries)... the original code was
```
        if (this.isStatsEnabled()) {
            CopyOnWriteArrayList<String> messageIds = getOrCreateMessageIds(executorName, teamId, methodName);
            Integer totalSize = messageIds.size();
            RateLimitQueue<SUPPLIER, MSG> queue = getRateLimitQueue(executorName, teamId);
            if (queue != null) {
                totalSize += queue.getCurrentActiveQueueSize(methodName);
            }
            getOrCreateTeamLiveStats(executorName, teamId).getCurrentQueueSize().put(methodName, totalSize);
            getOrCreateTeamLiveStats(executorName, teamId).getCurrentQueueSize().put(methodName, size);
```
Doesn't the second call `getOrCreateTeamLiveStats(executorName, teamId).getCurrentQueueSize().put(methodName, size);` reset the value to `size` (and make the calculation of `totalSize` not worth the cycles? In the case of the call from `execute`, `size` is always `0`.

seratch:
@gunrein Ah, you're right on the point. It seems the code should be improved, but this SDK has similar code in other modules too, so I will do more investigation on it later on. Thus, could you make this PR as simple as possible for now? Thank you so much for pointing the issue out.

gunrein:
@seratch - of course. The tier is now 3 and the other changes are reverted from the branch.

Thanks for taking a look at the issue.

seratch (approved):
LGTM

PR #1270, opened 2024-01-23
Fix #1269 A bug on the metric calculation inside async API clients

### LINKED ISSUE #1269: A bug on the metric calculation inside async API clients
As mentioned at https://github.com/slackapi/java-slack-sdk/pull/1268, the second line of code in this internal method seems to set an invalid value for the queue size: https://github.com/slackapi/java-slack-sdk/blob/v1.37.0/slack-api-client/src/main/java/com/slack/api/rate_limits/metrics/impl/BaseMemoryMetricsDatastore.java#L273-L274

The code should be corrected without any side effects to the current behavior.

seratch:
This pull request resolves #1269, thank you so much @gunrein !

The bug had been affecting only the output of the visible metrics data for developers. The internal smart rate limiter mechanism does not use the data at all, so the impact of the misbehavior should be relatively small.

## Requirements

### DIFF
--- a/slack-api-client/src/main/java/com/slack/api/audit/impl/AsyncRateLimitExecutor.java
+++ b/slack-api-client/src/main/java/com/slack/api/audit/impl/AsyncRateLimitExecutor.java
@@ -65,7 +65,7 @@ public class AsyncRateLimitExecutor {
         return CompletableFuture.supplyAsync(() -> {
             String messageId = messageIdGenerator.generate();
             addMessageId(teamId, methodName, messageId);
-            initCurrentQueueSizeStatsIfAbsent(teamId, methodName);
+            syncCurrentQueueSizeStats(teamId, methodName);
             if (NO_TOKEN_METHOD_NAMES.contains(methodName) || teamId == null) {
                 return runWithoutQueue(teamId, methodName, methodsSupplier);
             } else {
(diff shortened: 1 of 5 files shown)


slack-api-client/src/main/java/com/slack/api/audit/impl/AsyncRateLimitExecutor.java
```diff
@@ -80,9 +80,9 @@ public <T extends AuditApiResponse> CompletableFuture<T> execute(
         }, executorService);
     }
 
-    private void initCurrentQueueSizeStatsIfAbsent(String teamId, String methodNameWithSuffix) {
+    private void syncCurrentQueueSizeStats(String teamId, String methodNameWithSuffix) {
```
seratch, line 83: Since this method is a private one, we can safely rename it to represent more accurate behavior.

slack-api-client/src/main/java/com/slack/api/rate_limits/metrics/impl/BaseMemoryMetricsDatastore.java
```diff
@@ -264,13 +264,6 @@ public void updateCurrentQueueSize(String executorName, String teamId, String me
     @Override
     public void setCurrentQueueSize(String executorName, String teamId, String methodName, Integer size) {
         if (this.isStatsEnabled()) {
-            CopyOnWriteArrayList<String> messageIds = getOrCreateMessageIds(executorName, teamId, methodName);
```
seratch, line 267: These lines of code are completely unnecessary. They are just an overhead when executing this method.

PR #1325, opened 2024-06-16
Update MigrationExchange sample for Node SDK type

davidlj95:
In Slack's Node SDK, [there's an issue where `MigrationExchangeResponse` type is an interface instead of a map][issue]. Which forces consumers of the SDK to perform a cast in order to properly consume the API response.

Types for responses there are automagically generated [by a script](https://github.com/slackapi/node-slack-sdk/blob/%40slack/web-api%407.1.0/scripts/generate-web-api-types.sh)  based on the JSON samples in this repo. Then, script uses [`quicktype` package](https://www.npmjs.com/package/quicktype) in order to guess the type from the sample JSON. In order to tell `quicktype` that this field is a type, we can do a bit of trickery as @seratch [mentioned in the issue](https://github.com/slackapi/node-slack-sdk/issues/1584#issuecomment-1401230523). It's essentially using [specific property names](https://blog.quicktype.io/markov/) in examples in order to let `quicktype` infer the type is a map and not an interface & therefore avoid the cast by SDK consumers.

After a quick glance on how [`quicktype` infers what should be a map](https://github.com/glideapps/quicktype/blob/29bb8160660a7daa2a837ac0cb1e92f99729ad8e/packages/quicktype-core/src/rewrites/InferMaps.ts#L26-L35), seems the most straightforward way is to add at least 2 properties with digit-only numbers as keys. Sooo there you go.

Tested by running the types generation script checking out this branch and trick works. Will open a PR later in Node SDK after running script to bring the change there.

[issue]: https://github.com/slackapi/node-slack-sdk/issues/1584

## Requirements

### DIFF
--- a/json-logs/samples/api/migration.exchange.json
+++ b/json-logs/samples/api/migration.exchange.json
@@ -9,5 +9,9 @@
   "invalid_user_ids": [
     ""
   ],
-  "user_id_map": {}
+  "user_id_map": {
+    "0": "0",
+    "1": "1",
+    "2": "2"
+  }
 }


seratch (approved):
Thank you!

seratch:
@davidlj95 Thanks for sending this PR! I found that this needs more changes, thus I've made a squashed commit to including your commit credit: https://github.com/slackapi/java-slack-sdk/commit/33645edb3597c842ceca8e18a5894605bfd21a1b

PR #1326, opened 2024-06-18
add externalTeams.list sample json

filmaj:
Generated this using the API tester after connecting a production grid workspace to a production standalone workspace.


seratch:
Thanks, some details will be overwritten by automated tests later but there is no risk for merging this for now

filmaj:
I need to start adding the integration tests / the right way to do this!

PR #1337, opened 2024-07-16
Fix #1336 missing and invalid properties in rich text block elements

### LINKED ISSUE #1336: Rich Text Block Mismatch in Java Class and Json Version
We are trying to utilise Rich Text Block as Input , the user can add some placeholder ( we are using something like ${current}) and we would replace it with mentioning user.

We are facing issues wrt Rich Text Block and subsequent json format

1. In case the input contains the user mention, then then rich text type is "user". However, on converting to json, it shows as :

```json
{
  "type": "user",
  "user_id": "U03G3J3UGKU",
  "style": {
    "bold": true,
    "italic": false,
    "strike": false,
    "code": false
  }
}
```

Now, when we try to paste this in block kit builder or try sending this on UI, then it says "invalid additional property code".

On checking the commits :[ We think this commit is the issue](https://github.com/slackapi/java-slack-[sdk/blob/main/slack-api-model/src/main/java/com/slack/api/model/block/element/RichTextSectionElement.java) , since the  User object -> style object shouldn't have code attribute as mentioned here](https://api.slack.com/reference/block-kit/blocks#user-element-type)

Are there inconsistencies in Rich text block Java class and Rich text block  json version? 

### Reproducible in:

```xml
<dependency>
    <groupId>com.slack.api</groupId>
    <artifactId>bolt-servlet</artifactId>
    <version>1.39.0</version>
</dependency>
```

```bash
mvn dependency:tree | grep com.slack.api
gradle dependencies | grep com.slack.api
java -version 
sw_vers && uname -v # or `ver`
```

#### The Slack SDK version
```xml
<dependency>
    <groupId>com.slack.api</groupId>
    <artifactId>bolt-servlet</artifactId>
    <version>1.39.0</version>
</dependency>
```

(Paste the output of `mvn dependency:tree | grep com.slack.api` or `gradle dependencies | grep com.slack.api`)

#### Java Runtime version

(Paste the output of `java -version`)
java 11.0.16 2022-07-19 LTS
Java(TM) SE Runtime Environment 18.9 (build 11.0.16+11-LTS-199)
Java HotSpot(TM) 64-Bit Server VM 18.9 (build 11.0.16+11-LTS-199, mixed mode)
#### OS info
Mac Os
ProductName:	macOS
ProductVersion:	12.0.1
BuildVersion:	21A559
Darwin Kernel Version 21.1.0: Wed Oct 13 17:33:01 PDT 2021; root:xnu-8019.41.5~1/RELEASE_ARM64_T6000
#### Steps to reproduce:

(Share the commands to run, source code, and project settings (e.g., pom.xml/build.gradle))

1. Create a Rich text block input.
2. Add a user mention in the input
3. Read the user input using Slack bolt. -> You can see style properties being added like code, block etc.
4. Convert the user input to json 
5. Paste this in block kit builder and see error : invalid additional property code

#### Expected result:

```json
{
  "type": "user",
  "user_id": "U03G3J3UGKU",
  "style": {
    "bold": true,
    "italic": false,
    "strike": false,
    "code": false
  }
}
```

#### Actual result:
 error : invalid additional property code
 
 On checking the commits :[ We think this commit is the issue](https://github.com/slackapi/java-slack-[sdk/blob/main/slack-api-model/src/main/java/com/slack/api/model/block/element/RichTextSectionElement.java) , since the  User object -> style object shouldn't have code attribute as mentioned here](https://api.slack.com/reference/block-kit/blocks#user-element-type)
 
Screenshot 2024-07-15 at 8 45 34 PM

seratch:
This pull request resolves #1336 

## Requirements


slack-api-model/src/main/java/com/slack/api/model/block/element/RichTextSectionElement.java
```diff
@@ -42,7 +69,7 @@ public static class Channel implements RichTextElement {
         public static final String TYPE = "channel";
         private final String type = TYPE;
         private String channelId; // C12345678
-        private TextStyle style;
+        private NoCodeTextStyle style;
```
seratch, line 72: channel, user, usergroup, team, broadcast, color do not accept code property

slack-api-model/src/main/java/com/slack/api/model/block/element/RichTextSectionElement.java
```diff
@@ -146,7 +150,23 @@ public static class TextStyle {
         private boolean bold;
         private boolean italic;
         private boolean strike;
+        private boolean highlight;
```
seratch, line 153: these are not mentioned in the documents but actually these properties work for all elements

slack-api-model/src/main/java/com/slack/api/model/block/element/RichTextSectionElement.java
```diff
@@ -146,7 +150,23 @@ public static class TextStyle {
...
+        private boolean clientHighlight;
+        private boolean unlink;
         private boolean code;
     }
 
+    @Data
+    @Builder
+    @NoArgsConstructor
+    @AllArgsConstructor
+    public static class NoCodeTextStyle {
```
seratch, line 163: Added this for the ones that do not accept code

slack-api-model/src/test/java/test_locally/api/model/block/BlockKitTest.java
```diff
@@ -1667,4 +1667,144 @@ public void parseLinkTriggerMessages() {
         assertEquals(1, view.getAttachments().get(0).getBlocks().size());
         assertEquals(1, view.getBlocks().size());
     }
+
+
+    @Test
+    public void parseRichTextElements() {
+        String json = "{\n" +
+                "  \"blocks\": [\n" +
```
seratch, line 1675: Used this JSON data in Block Kit Builder for verifying

seratch (commented):
comments for reviewers

slack-api-model/src/main/java/com/slack/api/model/block/element/RichTextSectionElement.java
WilliamBergamin: It seems like this object may be referenced as `limited_style`, what do you think about naming this `LimitedTextStyle`?

```suggestion
    public static class LimitedTextStyle {
```

slack-api-model/src/main/java/com/slack/api/model/block/element/RichTextSectionElement.java
WilliamBergamin: 🤔 this makes sense, from what I was able to find there are currently only 2 versions of these style objects

WilliamBergamin (approved):
Great catches 💯 

Left one comment worth addressing but it is none-blocking

slack-api-model/src/main/java/com/slack/api/model/block/element/RichTextSectionElement.java
seratch: Sounds good. Renamed it.

Issue #919, opened 2022-01-26
Proposal: Separate example codes from project.


kuzuha:
I'm trying to migrate Java EE to Jakarta EE for some reason (e.g. native-image, spring 6).

This project has example code for many platform.
It's good for SDK users, but it includes some problems.

Now, some projects depends `javax` namespace, some other projects depends `jakarta` namespace.
(e.g. Jetty already migrated. Quarkus desided to stay `javax`)

So, I propose to separate example code from this project before Jakarta EE migration.
How do you think about this?

seratch:
Hi @kuzuha, thanks for writing in!

This is a great suggestion. 

As stated at https://github.com/slackapi/java-slack-sdk/issues/724, we were planning to completely migrate to Jakarta EE in the next major version (v2.x). 

However, while reading this issue, I came to think that this SDK should support both javax.servlet and jakarta.servlet as long as both are widely used in the communities.

Taking these into account, I'm leaning towards the following idea. We can add these new one in v1.x series as adding these are not a breaking change.
* Add `com.slack.api:bolt-jakarta-servlet` (jakarta.servlet 5.0), which is equivalent to `com.slack.api:bolt-servlet` (javax.servlet 3.1+)
* Add `com.slack.api:bolt-jakarta-jetty` (jetty 11), which is equivalent to `com.slack.api:bolt-jetty` (javax.servlet + jetty 9.x)

I'm not sure what Jakarta EE based solutions you use (or you're going to use) but the newly added `com.slack.api:bolt-jakarta-servlet` package should be useful for your use cases. Am I correct? Also, let me know if you have a different view/idea on this.

kuzuha:
Thanks for your response.

Supporting both EE is no pain for community.
I agree. But this project depends 2 `javax` namespace not only servlet.

expressly depending `javax.servlet-api`
* slack-app-backend
  * com.slack.api.app_backend.events.servlet
* bolt-servlet
  * com.slack.api.bolt.servlet

expressly depending `javax.websocket`
* bolt-socket-mode
  * included but not appear in code
* slack-api-client
  * com.slack.api.rtm
  * com.slack.api.socket-mode.impl

IMO, supporting both EE in one version is hard work.
The better way is keep mantainancing 1.x for a while.

seratch:
@kuzuha Thanks for your prompt reply!

[...]

PR #1352, opened 2024-08-22
Add Jakarta EE compatible Socket Mode client ref: #919

seratch:
This pull request adds two new optional modules:
* slack-jakarta-socket-mode-client
* bolt-jakarta-socket-mode

The currently available Socket Mode client's default implementation uses tyrus-standalone-client **1.x**, which is compatible with **javax.websocket-api** APIs. The Jakarta EE version of this interface is the **jakarta.websocket-client-api** APIs, and tyrus-standalone-client **2.x** is compatible with it. Since it's not feasible to have both tyrus-standalone-client 1.x and 2.x in the same module's dependencies, I have added a new module named slack-jakarta-socket-mode-client. See https://github.com/slackapi/java-slack-sdk/issues/919 for more details.

Developers can initialize the Jakarta-compatible SocketModeClient this way:

```java
import com.slack.api.Slack;
import com.slack.api.jakarta_socket_mode.JakartaSocketModeClientFactory;

public class Example {
    var appToken = System.getenv("SLACK_APP_TOKEN");
    var slack = Slack.getInstance();
    // Java EE compatible Socket Mode client
    slack.socketMode(appToken).connect();
    // Jakarta EE compatible Socket Mode client
    JakartaSocketModeClientFactory.create(slack, appToken).connect();
  }
}
```

In the same way, I’ve added a new Jakarta-compatible module, which is equivalent to bolt-socket-mode. Here is the demo code. As you can see, just replace the dependency and imports in the code:

```java
import com.slack.api.bolt.App;
import com.slack.api.bolt.jakarta_socket_mode.SocketModeApp;

public class Example {
    var app = new App();
    app.command("/hi", (req, ctx) -> {
      ctx.say("Hi there!");
      return ctx.ack();
    });
    var appToken = System.getenv("SLACK_APP_TOKEN");
    // Switch from com.slack.api.bolt.socket_mode to com.slack.api.bolt.jakarta_socket_mode
    new SocketModeApp(appToken, app).start();
  }
}
```

The reason behind this enhancement is that many Java-focused companies are planning to eliminate the legacy javax.* dependencies from their project settings. I don't think the short-term risk of having a javax.websocket dependency is significant, but it seems it's about time to provide an option for migration on the developers' side.

## Requirements


slack-jakarta-socket-mode-client/src/main/java/com/slack/api/jakarta_socket_mode/impl/JakartaSocketModeClientTyrusImpl.java
```diff
@@ -0,0 +1,435 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Socket Mode Client
```
WilliamBergamin, line 36:
```suggestion
 * Jakarta Socket Mode Client
```

slack-jakarta-socket-mode-client/src/main/java/com/slack/api/jakarta_socket_mode/impl/JakartaSocketModeClientTyrusImpl.java
```diff
@@ -0,0 +1,435 @@
+package com.slack.api.jakarta_socket_mode.impl;
```
WilliamBergamin, line 1: I see this is almost an exact copy of `SocketModeClientTyrusImpl.java`, which is production worthy 🧠

WilliamBergamin (approved):
This looks good to me 💯  the changes seem to be mainly be a copy of existing code 👍 

I'm assuming this code has been stable for a while, maintaining this type of duplication should not be a problem

slack-jakarta-socket-mode-client/src/main/java/com/slack/api/jakarta_socket_mode/impl/JakartaSocketModeClientTyrusImpl.java
seratch: Yeah, you're right!

PR #1355, opened 2024-08-30
Updating app_requested payload to include new field

gugutagli:
Updating ```app_requested``` payload to include ```is_granular_bot_app```

* [x] **slack-api-*-kotlin-extension** (Kotlin Extensions for Slack API Clients)

## Requirements

### DIFF
--- a/slack-api-model/src/main/java/com/slack/api/model/admin/App.java
+++ b/slack-api-model/src/main/java/com/slack/api/model/admin/App.java
@@ -12,11 +12,12 @@ public class App {
     private String privacyPolicyUrl;
     private String appHomepageUrl;
     private String appDirectoryUrl;
-    @SerializedName("is_granular_bot_app")
     @SerializedName("is_app_directory_approved")
     private boolean appDirectoryApproved;
     @SerializedName("is_internal")
     private boolean internal;
+    @SerializedName("is_granular_bot_app")
+    private boolean granularBotApp;
     private String additionalInfo;
     private AppIcons icons;
 }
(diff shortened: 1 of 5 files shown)


seratch (changes requested):
This Java class needs to be updated and you can run the unit test that auto-generates the sample JSON file: https://github.com/slackapi/java-slack-sdk/blob/main/slack-api-model/src/main/java/com/slack/api/model/admin/AppRequest.java

gugutagli:
> This Java class needs to be updated and you can run the unit test that auto-generates the sample JSON file: https://github.com/slackapi/java-slack-sdk/blob/main/slack-api-model/src/main/java/com/slack/api/model/admin/AppRequest.java

Thank you, I was able to add to the java file. This is my first time working on this, which test do I run and how do I run them?

slack-api-model/src/main/java/com/slack/api/model/admin/App.java
```diff
@@ -12,6 +12,7 @@ public class App {
     private String privacyPolicyUrl;
     private String appHomepageUrl;
     private String appDirectoryUrl;
+    @SerializedName("is_granular_bot_app")
```
seratch, line 15:
```suggestion
    @SerializedName("is_granular_bot_app")
    private boolean granularBotApp
```

seratch (changes requested):

seratch:
@gugutagli You can run [this script](https://github.com/slackapi/java-slack-sdk/blob/main/scripts/run_no_prep_tests.sh) for it, but it may take time. For this specific pull request, running only [this test](https://github.com/slackapi/java-slack-sdk/blob/main/slack-app-backend/src/test/java/test_locally/sample_json_generation/EventsApiPayloadDumpTest.java) should be fine.

slack-api-model/src/main/java/com/slack/api/model/admin/App.java
seratch: If this flag can be absent in some patterns, `private Boolean isGranularBotApp` without SerializedName annotation is suitable.

seratch:
@gugutagli Thanks for sending this PR!

PR #1403, opened 2024-12-05
Fix #1402 Migrating AWS S3 SDK to v2

### LINKED ISSUE #1402: Migrating AWS S3 SDK to v2
As reported at https://github.com/slackapi/java-slack-sdk/pull/1168#pullrequestreview-2474523299, our optional dependency AWS S3 client SDK must be migraded to its v2 by the end of 2025.

Both groupId and artifactId are different:
```diff
-            <groupId>com.amazonaws</groupId>
-            <artifactId>aws-java-sdk-s3</artifactId>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>s3</artifactId>
```

Here is a version range example:
```diff
-        <aws.s3.version>[1.12.62,1.13.0)</aws.s3.version>
+        <aws.s3.version>[2.29.24,3)</aws.s3.version>
```

V2 is fully source-code incompatible, so this work requires rewriting the modules relying on the AWS SDK plus testing the behavior is still the same. It seems there is no need to migrate "aws-lambda-java-core" to something new this time.

See also:
- https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/migration.html
- https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/migration-service-changes.html
- https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/migration-client-credentials.html
- https://docs.aws.amazon.com/lambda/latest/dg/lambda-java.html

We may want to try the migration tool if it works well for us:
- https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/migration-tool.html

### Requirements

Please make sure if this topic is specific to this SDK. For general questions/issues about Slack API platform or its server-side, could you submit questions at https://my.slack.com/help/requests/new instead. :bow:

seratch:
This pull request resolves #1402

## Requirements

### DIFF
--- a/bolt/src/main/java/com/slack/api/bolt/service/builtin/AmazonS3InstallationService.java
+++ b/bolt/src/main/java/com/slack/api/bolt/service/builtin/AmazonS3InstallationService.java
@@ -1,11 +1,5 @@
 package com.slack.api.bolt.service.builtin;
 
-import com.amazonaws.auth.AWSCredentials;
-import com.amazonaws.auth.DefaultAWSCredentialsProviderChain;
-import com.amazonaws.services.s3.AmazonS3;
-import com.amazonaws.services.s3.AmazonS3ClientBuilder;
-import com.amazonaws.services.s3.model.*;
-import com.amazonaws.util.IOUtils;
 import com.slack.api.bolt.Initializer;
 import com.slack.api.bolt.model.Bot;
 import com.slack.api.bolt.model.Installer;
@@ -14,14 +8,25 @@ import com.slack.api.bolt.model.builtin.DefaultInstaller;
 import com.slack.api.bolt.service.InstallationService;
 import com.slack.api.bolt.util.JsonOps;
 import lombok.extern.slf4j.Slf4j;
+import software.amazon.awssdk.auth.credentials.AwsCredentials;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
+import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;
+import software.amazon.awssdk.core.ResponseBytes;
+import software.amazon.awssdk.core.sync.RequestBody;
+import software.amazon.awssdk.core.sync.ResponseTransformer;
+import software.amazon.awssdk.services.s3.S3Client;
+import software.amazon.awssdk.services.s3.model.*;
 
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.Map;
 import java.util.Optional;
 
 @Slf4j
 public class AmazonS3InstallationService implements InstallationService {
 
     private final String bucketName;
+    private AwsCredentialsProvider credentialsProvider;
     private boolean historicalDataEnabled;
 
     public AmazonS3InstallationService(String bucketName) {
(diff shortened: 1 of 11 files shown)


bolt/src/main/java/com/slack/api/bolt/service/builtin/AmazonS3OAuthStateService.java
```diff
@@ -33,74 +33,91 @@ public AmazonS3OAuthStateService(String bucketName) {
...
                 throw new IllegalStateException("AWS credentials not found");
             }
             if (log.isDebugEnabled()) {
-                log.debug("AWS credentials loaded (access key id: {})", credentials.getAWSAccessKeyId());
+                log.debug("AWS credentials loaded (access key id: {})", credentials.accessKeyId());
+            }
+            boolean bucketExists = false;
+            Exception error = null;
+            try (S3Client s3 = createS3Client()) {
+                bucketExists = s3.headBucket(HeadBucketRequest.builder().bucket(bucketName).build()) != null;
```
acharyashashank, line 47: @seratch San: Do you think that [this](https://docs.aws.amazon.com/ja_jp/AmazonS3/latest/API/s3_example_s3_Scenario_DoesBucketExist_section.html) implementation from the AWS is better than then using ``HeadBucketRequest``

seratch: @acharyashashank Thanks for the suggestion! 

I switched to headBucket because the AWS migration guide recommends using the API in version 2: https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/migration-service-changes.html. Do you think using getBucketAcl instead could bring benefits such as faster runtime performance?

Even if I use the code snippet provided there, I won't follow AwsServiceException error handling because the goal of this code is to verify if the client has permissions to fetch/modify S3 bucket data, rather than checking the bucket's existence.

bolt/src/main/java/com/slack/api/bolt/service/builtin/AmazonS3InstallationService.java
```diff
@@ -32,16 +37,24 @@ public AmazonS3InstallationService(String bucketName) {
...
+            Exception error = null;
+            try (S3Client s3 = createS3Client()) {
+                bucketExists = s3.headBucket(HeadBucketRequest.builder().bucket(bucketName).build()) != null;
+            } catch (Exception e) { // NoSuchBucketException etc.
+                error = e;
             }
-            boolean bucketExists = createS3Client().doesBucketExistV2(bucketName);
             if (!bucketExists) {
-                throw new IllegalStateException("Failed to access the Amazon S3 bucket (name: " + bucketName + ")");
+                String message = "Failed to access the Amazon S3 bucket (name: " + bucketName + ", error: " + error + ")";
```
WilliamBergamin, line 56: Will this print out only the `error` message or the message with the stack trace?

Do we want to have the message of this error contain the trace of another error 🤔

WilliamBergamin (approved):
I'm a bit rusty on my java but this looks good to me 👍 

Left once comment about error messages, but should not be a blocker

bolt/src/main/java/com/slack/api/bolt/service/builtin/AmazonS3InstallationService.java
seratch: This prints only this exception, but it depends on how toString method is implemented by an exception class. So, I've tweaked the code to have only the class name and message. Thanks for the suggestion!

PR #1411, opened 2025-01-07
Fix #1409 Socket Mode: Slow message consumption when listeners do not immediately return ack()

### LINKED ISSUE #1409: Socket Mode: Slow message consumption when listeners do not immediately return ack()
(Describe your issue and goal here)

* [x] **bolt-jakarta-socket-mode** (Bolt for Java - optional modules)

### Requirements

We are using java sdk in socket mode with tyrus client.
I just found out that it runs message processing single threaded and:
1. it looks like it's not really intended.
com.slack.api.socket_mode.SocketModeClient#initializeMessageProcessorExecutor
It creates 10 (by default) jobs for processing messages with executor service created by the com.slack.api.util.thread.DaemonThreadExecutorServiceProvider#createThreadScheduledExecutor
which creates a single threaded executor
I've tried to test it by putting Thread.sleep for a couple of seconds in event processor and spamming messages and the message processing was very slow which proves the point.
It looks like a bug to me.
2. It's hard to configure concurrency – it's configured in the clients constructor, the client is created via JakartaSocketModeClientFactory#create(Slack, String) method which is called in private method SocketModeApp#buildSocketModeClientFactory that doesn't use any top-level properties that can overwrite the behaviour.
It would be nice to have an easy option to change concurrency.
3. It uses scheduled executor which significantly reduces the number of ExecutorServices that is possible to use for message processing. Is there a reason for using scheduled instead of a regular one? From my perspective switching to a regular one should simplify things, because it already encapsulates the queueing logic. + it would allow to use virtual threads executor
If the used executor type were changed and the ability to provide the concrete implementation were given, it would mitigate the second problem in this list from my perspective

Let me know if I can clarify anything else

seratch:
This pull request resolves #1409 by adding a new "concurrency" option to the bolt-socket-mode and bolt-jakarta-socket-mode libraries.

The current implementation delays picking a new queued WebSocket message until the preceding operation returns an ack() WebSocket message to Slack. For context, we recommend that developers move all time-consuming tasks to `app.executorService().execute` (or a similar solution). If an app is built this way, there should be no issue with this mechanism. However, if a listener code takes a long time for a particular scenario, this situation can negatively impact the runtime performance of subsequent event handling. You can reproduce the issue with the following code snippet:

```java
app.event(MessageEvent.class, (req, ctx) -> {
    // Without concurrency option, this time-consuming task slows the whole message processing mechanism down
    try {
        Thread.sleep(1000L);
    } catch (InterruptedException e) {
        throw new RuntimeException(e);
    }
    ctx.asyncClient().reactionsAdd(r -> r
            .channel(req.getEvent().getChannel())
            .name("eyes")
            .timestamp(req.getEvent().getTs())
    );
    return ctx.ack();
});

String appToken = System.getenv(Constants.SLACK_SDK_TEST_SOCKET_MODE_APP_TOKEN);
SocketModeApp socketModeApp = new SocketModeApp(appToken, app);
socketModeApp.start();
```

With the new concurrency option, Bolt apps can resiliently handle events at a constant pace (as long as there are available threads in the thread pool).

```java
// SocketModeApp socketModeApp = new SocketModeApp(appToken, app);
SocketModeApp socketModeApp = new SocketModeApp(appToken, app, 10);
```

We won't change the default behavior this time, but it's worth considering enabling concurrency by default in future releases.

## Requirements

### DIFF
--- a/bolt-jakarta-socket-mode/src/main/java/com/slack/api/bolt/jakarta_socket_mode/SocketModeApp.java
+++ b/bolt-jakarta-socket-mode/src/main/java/com/slack/api/bolt/jakarta_socket_mode/SocketModeApp.java
@@ -3,14 +3,15 @@ package com.slack.api.bolt.jakarta_socket_mode;
 import com.google.gson.Gson;
 import com.google.gson.JsonElement;
 import com.slack.api.bolt.App;
-import com.slack.api.bolt.request.Request;
-import com.slack.api.bolt.response.Response;
 import com.slack.api.bolt.jakarta_socket_mode.request.SocketModeRequest;
 import com.slack.api.bolt.jakarta_socket_mode.request.SocketModeRequestParser;
+import com.slack.api.bolt.request.Request;
+import com.slack.api.bolt.response.Response;
 import com.slack.api.jakarta_socket_mode.JakartaSocketModeClientFactory;
 import com.slack.api.socket_mode.SocketModeClient;
 import com.slack.api.socket_mode.response.AckResponse;
 import com.slack.api.util.json.GsonFactory;
+import com.slack.api.util.thread.DaemonThreadExecutorServiceFactory;
 import lombok.Builder;
 import lombok.Data;
 import lombok.extern.slf4j.Slf4j;
@@ -18,6 +19,7 @@ import lombok.extern.slf4j.Slf4j;
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ExecutorService;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
@@ -27,6 +29,7 @@ public class SocketModeApp {
     private final App app;
     private final Supplier<SocketModeClient> clientFactory;
     private SocketModeClient client;
+    private final ExecutorService executorService;
 
     private static final Function<ErrorContext, Response> DEFAULT_ERROR_HANDLER = (context) -> {
         Exception e = context.getException();
@@ -69,7 +72,8 @@ public class SocketModeApp {
     private static Supplier<SocketModeClient> buildSocketModeClientFactory(
             App app,
             String appToken,
-            Function<ErrorContext, Response> errorHandler
+            Function<ErrorContext, Response> errorHandler,
+            ExecutorService executorService
     ) {
         return () -> {
             try {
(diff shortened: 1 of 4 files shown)


bolt-socket-mode/src/main/java/com/slack/api/bolt/socket_mode/SocketModeApp.java
```diff
@@ -147,12 +203,43 @@ public SocketModeApp(
...
+            ExecutorService executorService
+    ) throws IOException {
+        this(buildSocketModeClientFactory(app, appToken, backend, errorHandler, executorService), app, executorService);
+    }
+
+    // intentionally private to avoid exposing the ExecutorService initialization
+    private SocketModeApp(
+            Supplier<SocketModeClient> clientFactory,
+            App app,
+            ExecutorService executorService
```
Mugenor, line 238: Is there a reason for hiding ExecutorService initialisation?
IMO, it would be nice to be able to provide any type of executor service

Mugenor (changes requested):

bolt-socket-mode/src/main/java/com/slack/api/bolt/socket_mode/SocketModeApp.java
seratch: We encourage making a daemon thread one like this code does. Are you planning to use a different one?

bolt-jakarta-socket-mode/src/main/java/com/slack/api/bolt/jakarta_socket_mode/SocketModeApp.java
```diff
@@ -69,35 +72,22 @@ private static void sendSocketModeResponse(
...
-                            long spentMillis = System.currentTimeMillis() - startMillis;
-                            log.debug("Response time: {} milliseconds", spentMillis);
-                        }
+                    if (executorService != null) {
+                        // asynchronous
+                        executorService.execute(() -> runBoltApp(
+                                message, app, client, requestParser, errorHandler, gson));
+                    } else {
+                        // synchronous
+                        runBoltApp(message, app, client, requestParser, errorHandler, gson);
```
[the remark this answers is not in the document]
seratch: Thanks for the suggestion! Indeed, this approach is also an option, but for now I'd like to avoid moving the runBoltApp to a different thread, even though it's a slight difference. When we make the concurrent mode the default, simplifying the code as you suggest is worth considering.

bolt-socket-mode/src/main/java/com/slack/api/bolt/socket_mode/SocketModeApp.java
seratch: OK, I'd like to keep this parameter hidden from developers for now, but we may consider accepting ExecutorService in future updates. Thanks for your comment!

Issue #1452, opened 2025-03-12
MethodsCompletionException is unnecessarily complex


osigida:
The [com.slack.api.methods.MethodsCompletionException](https://github.com/slackapi/java-slack-sdk/blob/main/slack-api-client/src/main/java/com/slack/api/methods/MethodsCompletionException.java) does not align with the partner's API (getMessage) and is challenging to maintain.  It requires handling specific exceptions and examining complex details to retrieve information. 
However, the implementation is unnecessarily complex and could be significantly simplified.

#### The Slack SDK version
1.45.3

#### Expected result:
It is easy to use Slack exceptions because they follow the default behavior and implement the default methods

#### Actual result:
Exception handling requires some specialized code

WilliamBergamin:
Hi @osigida thanks for writing is 💯 

This seems like a good feature request, this may involve a breaking change and in this case we would need to include it in our next major release

PR #1453, opened 2025-03-12
[1452] simplify MethodsCompletionException implementation

osigida:
MethodsCompletionException is simplified

* [X] **slack-api-client** (Slack API Clients)

### DIFF
--- a/slack-api-client/src/main/java/com/slack/api/methods/MethodsCompletionException.java
+++ b/slack-api-client/src/main/java/com/slack/api/methods/MethodsCompletionException.java
@@ -1,27 +1,13 @@
 package com.slack.api.methods;
 
-import lombok.Data;
 import lombok.EqualsAndHashCode;
-import lombok.extern.slf4j.Slf4j;
-
-import java.io.IOException;
 
 /**
  * Wrapped exception holding the cause exception occurred in AsyncMethodsClient
  */
-@Data
-@Slf4j
-@EqualsAndHashCode(callSuper = false)
+@EqualsAndHashCode(callSuper = true)
 public class MethodsCompletionException extends RuntimeException {
-
-    private final IOException ioException;
-    private final SlackApiException slackApiException;
-    private final Exception otherException;
-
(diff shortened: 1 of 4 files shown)


WilliamBergamin (changes requested):
Hi @osigida thanks for your interest and taking time to open a PR for this 💯 

Unfortunately these changes would break the current interface contract established by by the library. We would consider this a breaking change and would need to include it as part of a major version.

We can include these types of improvements in our next major version but for the time being we do not have plans to release one in the near future.

I would suggest closing this PR for the time being and keeping #1452 open in order to track the feature request

osigida:
Hi @WilliamBergamin! thanks a lot for the reply. 
I can make it backward compatible — just a few methods to support previous implementation. 
And backward compatibility could be removed in 2.0
Should I?

slack-api-client/src/main/java/com/slack/api/methods/MethodsCompletionException.java
```diff
@@ -1,27 +1,13 @@
...
 /**
  * Wrapped exception holding the cause exception occurred in AsyncMethodsClient
  */
-@Data
-@Slf4j
-@EqualsAndHashCode(callSuper = false)
+@EqualsAndHashCode(callSuper = true)
 public class MethodsCompletionException extends RuntimeException {
-
-    private final IOException ioException;
```
seratch, line 17: Let me share my thoughts as former maintainer of this project:

The underlying API client (MethodsClient) can throw any of these, and the current implementation provides getters to access them. So, it's not because of the breaking change, these data structure should not be eliminated. I hear that having three types here could be unnecessary for most cases, but this could be still beneficial for other users.

If your concern is the lack of easy-to-use getMessage() method, the internal code could be changed just to override getMessage() to return the underlying exception's message string data. Only one of these three usually exists, so checking the existence of them one by one within overridden getMessage() method should work for the purpose.

This could be a minor breaking change, so ideally it should be considered in the next major version. But in practice, having such a change with sufficient notice in release notes should be okay too.

PR #1472, opened 2025-05-09
Replace misspellings of Jakarta in reference documentation

rbatsenko:
jakarta name fix


rbatsenko:
Already signed, not sure why it says I didn't.

zimeg (approved):
@rbatsenko Thank you so much for catching this and sharing a fix 📚 ✨ 

This change matches the download on [Maven](https://central.sonatype.com/search?q=g:com.slack.api%20%20a:bolt-jakarta-socket-mode&smo=true) and the package path within this project so let's now merge this!

PR #1510, opened 2025-10-01
feat: Centralize and expose a registerTypeAdapters method on GsonFactory

luanpotter:
Centralize and expose a `registerTypeAdapters` method on `GsonFactory`

Typically on larger projects, `gson` needs to be configured with many other adapters and configurations. The way the `GsonFactory` class on `slack-api-client` currently expose a finalized, pre-built `gson` instances makes it less reusable for users, who would have to copy the adapter registration configuration one-by-one if they already have a `gson` of their own.

In fact, even within `GsonFactory`, the code is already repeated thrice, which can lead to desynchronization issues - which, in fact, did happen, as the last version had two repeated registrations (`MessageChangedEvent.PreviousMessage` and `LogsResponse.DetailsChangedValue`) and two missing registrations (`FunctionExecutedEvent.InputValue`, `Attachment.VideoHtml`), clearly a copy-paste mistake (unless this was intentional, but the reason would most certainly elude me).

This unifies all instances to use the same shared method, and expose it so users can create their own instances as well.

## Requirements

### DIFF
--- a/.github/workflows/unit-tests-jdk-14.yml
+++ b/.github/workflows/unit-tests-jdk-14.yml
@@ -30,6 +30,6 @@ jobs:
         env:
           SKIP_UNSTABLE_TESTS: 1
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@18283e04ce6e62d37312384ff67231eb8fd56d24 # v5.4.3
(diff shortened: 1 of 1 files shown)


luanpotter:
@WilliamBergamin my pleasure! and thank you so much for the swift review 🙏

PR #1512, opened 2025-10-07
Add work objects support

vegeris:
(Describe the goal of this PR. Mention any related Issue numbers)

## Requirements


docs/japanese/guides/assistants.md
```diff
@@ -105,7 +105,7 @@ app.blockAction("assistant-generate-numbers", (req, ctx) -> {
         .channel(req.getPayload().getChannel().getId())
         .threadTs(req.getPayload().getMessage().getThreadTs())
         .text("OK, I will generate numbers for you!")
-        .metadata(new Message.Metadata("assistant-generate-numbers", eventPayload))
+        .metadata(Message.Metadata.builder().eventType("assistant-generate-numbers").eventPayload(eventPayload).build())
```
vegeris: Similar to node-slack-sdk, I originally updated the Message.Metadata object that was previously event-metadata-only. Planning on updating this PR to use a separate EntityAndEventMessageMetadata object as well

cc: @zimeg and @mwbrooks to sanity check this; any apps calling chat.postMessage and constructing the Message.Metadata object will need to update that type to Message.EntityAndEventMessageMetadata

slack-api-client/src/main/java/com/slack/api/methods/MethodsClient.java
```diff
@@ -908,30 +1173,40 @@ AdminUsergroupsRemoveChannelsResponse adminUsergroupsRemoveChannels(
     // Developer preview has ended
     // This feature was exclusive to our workspace apps developer preview.
     // The preview has now ended, but fan-favorite features such as token rotation
-    // and the Conversations API will become available to classic Slack apps over the coming months.
+    // and the Conversations API will become available to classic Slack apps over
+    // the coming months.
```
zimeg, line 1177: 🪬 note: TIL!

slack-api-client/src/main/java/com/slack/api/methods/MethodsClient.java
```diff
@@ -1494,616 +1958,781 @@ AdminUsergroupsRemoveChannelsResponse adminUsergroupsRemoveChannels(
...
 
-    WorkflowsStepFailedResponse workflowsStepFailed(RequestConfigurator<WorkflowsStepFailedRequest.WorkflowsStepFailedRequestBuilder> req) throws IOException, SlackApiException;
+    WorkflowsUpdateStepResponse workflowsUpdateStep(
+            RequestConfigurator<WorkflowsUpdateStepRequest.WorkflowsUpdateStepRequestBuilder> req)
+            throws IOException, SlackApiException;
 
-    WorkflowsUpdateStepResponse workflowsUpdateStep(WorkflowsUpdateStepRequest req) throws IOException, SlackApiException;
+    // ------------------------------
+    // work object entities
+    // ------------------------------
```
zimeg: 🧮 quibble: This method might've been renamed in recent times. Can I request a realphabeticalization before merge?

slack-api-client/src/main/java/com/slack/api/methods/Methods.java
```diff
@@ -725,4 +731,10 @@ private Methods() {
     public static final String WORKFLOWS_STEP_COMPLETED = "workflows.stepCompleted";
     public static final String WORKFLOWS_STEP_FAILED = "workflows.stepFailed";
     public static final String WORKFLOWS_UPDATE_STEP = "workflows.updateStep";
+
+    // ------------------------------
+    // work object entities
+    // ------------------------------
+
+    public static final String ENTITY_PRESENT_DETAILSs = "entity.presentDetails";
```
zimeg:
```suggestion
```

🪓 suggestion(blocking): This might be alright to remove in favor of less plural before?

slack-api-client/src/main/java/com/slack/api/methods/AsyncMethodsClient.java
```diff
@@ -244,6 +246,7 @@
 import com.slack.api.methods.response.workflows.WorkflowsStepFailedResponse;
 import com.slack.api.methods.response.workflows.WorkflowsUpdateStepResponse;
 
+import java.io.IOException;
```
zimeg:
```suggestion
```

🪓 question: Is this safe to remove?

slack-api-client/src/main/java/com/slack/api/methods/AsyncMethodsClient.java
```diff
@@ -1565,4 +1568,12 @@ CompletableFuture<AdminConversationsWhitelistListGroupsLinkedToChannelResponse>
 
     CompletableFuture<WorkflowsUpdateStepResponse> workflowsUpdateStep(RequestConfigurator<WorkflowsUpdateStepRequest.WorkflowsUpdateStepRequestBuilder> req);
 
+    // ------------------------------
+    // work object entities
```
zimeg: 👁️‍🗨️ nit: Kind request to trail `emoji` methods for this section too?

slack-app-backend/src/main/java/com/slack/api/app_backend/events/handler/EntityDetailsRequestedHandler.java
```diff
@@ -0,0 +1,14 @@
+
```
zimeg:
```suggestion
```

🪓 nit: Perhaps this might not break imports but for fast reads-

slack-api-client/src/test/java/test_locally/api/methods_admin_api/FieldValidationTest.java
```diff
@@ -442,12 +442,12 @@ public void adminEmoji() throws Exception {
         {
             AdminEmojiAddResponse obj = parse(prefix + "add", AdminEmojiAddResponse.class);
             verifyIfAllGettersReturnNonNull(obj);
-            verifyIfAllGettersReturnNonNullRecursively(obj.getResponseMetadata(), "getMessages");
+            //verifyIfAllGettersReturnNonNullRecursively(obj.getResponseMetadata(), "getMessages"); // TODO
```
zimeg: 🎯 question: Can this be reverted without breaking tests?

slack-api-client/src/main/java/com/slack/api/methods/response/chat/ChatUnfurlResponse.java
```diff
@@ -15,5 +15,6 @@ public class ChatUnfurlResponse implements SlackApiTextResponse {
     private String needed;
     private String provided;
     private transient Map<String, List<String>> httpResponseHeaders;
-
+    private String callstack;
+    private Object response_metadata;
```
zimeg:
```java
import com.slack.api.model.ErrorResponseMetadata;
```

```suggestion
    private ErrorResponseMetadata responseMetadata;
```

🐫 suggestion: I believe we can **camelCase** attributes in code with deserialization matching the API?

⚠️ question: Is this also `response_metadata` returned with an error? I found a similar value with `chat.postMessage` that motivated this suggestion!

slack-api-client/src/main/java/com/slack/api/methods/impl/MethodsClientImpl.java
```diff
@@ -3566,6 +3568,16 @@ public WorkflowsUpdateStepResponse workflowsUpdateStep(RequestConfigurator<Workf
...
 
+    @Override
+    public EntityPresentDetailsResponse entityPresentDetails(EntityPresentDetailsRequest req) throws IOException, SlackApiException {
+        return postFormWithTokenAndParseResponse(toForm(req), Methods.ENTITY_PRESENT_DETAILSs, getToken(req), EntityPresentDetailsResponse.class);
+    }
+
+    @Override
+    public EntityPresentDetailsResponse entityPresentDetails(RequestConfigurator<EntityPresentDetailsRequest.EntityPresentDetailsRequestBuilder> req) throws IOException, SlackApiException {
+        return entityPresentDetails(req.configure(EntityPresentDetailsRequest.builder()).build());
+    }
```
zimeg: 🤖 quibble: The ordering of this one caught me off guard! Can I kind request following `emoji` with these?

slack-api-client/src/main/java/com/slack/api/methods/impl/AsyncMethodsClientImpl.java
```diff
@@ -2873,4 +2875,14 @@ public CompletableFuture<WorkflowsUpdateStepResponse> workflowsUpdateStep(Reques
...
 
+    @Override
+    public CompletableFuture<EntityPresentDetailsResponse> entityPresentDetails(EntityPresentDetailsRequest req) {
+        return executor.execute(ENTITY_PRESENT_DETAILS, toMap(req), () -> methods.entityPresentDetails(req));
+    }
+
+    @Override
+    public CompletableFuture<EntityPresentDetailsResponse> entityPresentDetails(RequestConfigurator<EntityPresentDetailsRequest.EntityPresentDetailsRequestBuilder> req) {
+        return entityPresentDetails(req.configure(EntityPresentDetailsRequest.builder()).build());
+    }
```
zimeg: 📠 nit: Personal preference for this section to move up a few lines for alphabetics-

slack-api-client/src/main/java/com/slack/api/methods/request/chat/ChatUnfurlRequest.java
```diff
@@ -75,6 +76,17 @@ public class ChatUnfurlRequest implements SlackApiRequest {
     // https://docs.slack.dev/changelog/2021-08-changes-to-unfurls
     private String source;
 
+    private String rawMetadata;
```
zimeg, line 82:
```suggestion
    private String metadataAsString;
```

🔍 suggestion: To match `chat.postMessage` formats!

https://github.com/slackapi/java-slack-sdk/blob/e3f79e0e2512d5be5cf1248327b9ca07580e8cf9/slack-api-client/src/main/java/com/slack/api/methods/request/chat/ChatPostMessageRequest.java#L66

zimeg: 📚 suggestion: Adding similar **javadoc** might be nice too?

docs/japanese/guides/assistants.md
zimeg: 🤔 @vegeris This is so interesting! Thank you for calling it out 🙏 ✨ 

If I'm understanding this right, changing the `metadata` type from `Message.Metadata` might error during compilation of app code for this:

```java
Message.Metadata metadata = message.getMetadata();
```

It makes sense that this argument can be either metadata type for the API but I'm wondering if we workaround changing this in the SDK for now with an additional argument for the method input and generous javadoc too?:

```java
/**
 * Metadata of entities attached to a message. Used instead of the "metadata" attribute, which sends event-based message metadata. In an upcoming major version, the "metadata" attribute will accept both types of metadata.
 */ 
private Message.EntityMessageMetadata entityMessageMetadata;
```

IIRC we have logic to parse various cases of `metadata` that can be extended in meantimes-

slack-api-client/src/main/java/com/slack/api/methods/RequestFormBuilder.java
```diff
@@ -2999,15 +3029,36 @@ public static FormBody.Builder toForm(WorkflowsUpdateStepRequest req) {
         return form;
     }
 
+    public static FormBody.Builder toForm(EntityPresentDetailsRequest req) {
```
zimeg: 🔐 quibble: Can these parsings be alphabetized earlier in the file?

slack-api-client/src/main/java/com/slack/api/methods/request/entity/EntityPresentDetailsRequest.java
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * Authentication token.
+     */
+    private String token;
+
+    /**
+     * JSON object containing the data that will be displayed in the flexpane for
+     * the entity.
+     */
+    private String rawMetadata;
```
zimeg, line 26:
```suggestion
    private String metadataAsString;
```

🧵 suggestion: To match another comment, but I also think duplicating the **javadoc** for the more explicit typings that follow might be solid too for reference generation? I haven't experimented with such edge-

slack-api-model/src/main/java/com/slack/api/model/Attachment.java
```diff
@@ -357,6 +357,7 @@ public static class Record {
     private String mimetype;
     private String url;
     private AttachmentMetadata metadata;
+    private Object workObjectEntity;
```
zimeg: 👾 question: Is this particular to a specific `Object` import, or can it be cast to something else? IIRC adjustments to these models are common if fields are adjusted, but having such type might make accessing easier?

zimeg (commented):
🌟 @vegeris Praises all throughout these changes! I tried to keep comments focused to thoughts and suggestions. This is looking solid!

The `metadata` typings you've called out caught my attention and I left some rambles on a workaround to avoid changing the existing types. I'm not sure if it's a good idea but I wonder if some changes might be left as `TODO` for now?

So curious about what you think! I plan to test this more after updates to metadata typings that might be in progress at the moment too 🫡

slack-api-client/src/test/java/test_with_remote_apis/methods/chat_Test.java
```diff
@@ -43,1026 +47,1288 @@
...
+                        if (ts != null) {
+                                final String _ts = ts;
+                                slack.methods(botToken).chatDelete(r -> r.channel(randomChannelId).ts(_ts));
+                        }
                 }
-            }
         }
-    }
-
-    String blocksAsString = "[\n" +
```
vegeris, line 79: Looks like my formatter in VSCode updated this whole file; is there a particular formatter I should be using? (My current default is `Language Support for Java(TM) by Red Hat`)

slack-api-model/src/main/java/com/slack/api/model/Attachment.java
vegeris: The work object attachment is the internal client schema, which we don't want users to consume, but in the integration test for `chat.unfurl` I'm verifying that the work object was successfully added by checking for this property

docs/japanese/guides/assistants.md
vegeris: Thanks for looking into this!

> additional argument for the method input

I'm not sure I follow 👀 
Edit: Ohh, gotcha; potentially have a separate property for entity+event metadata and check for it in `FormBody.Builder toForm(ChatPostMessageRequest req)`

I believe @WilliamBergamin is also back this week; it would be great to also get your input 🙏

slack-api-client/src/main/java/com/slack/api/methods/Methods.java
vegeris: Whoops; got mixed up and added it twice!

slack-api-client/src/main/java/com/slack/api/methods/MethodsClient.java
vegeris: Will do; 'work objects' was the internal name which was potentially going to change (so we were avoiding using it in external API names etc.) but we ended up keeping it

Here I'll just use 'entity' as the header to be consistent with the API naming

slack-api-client/src/main/java/com/slack/api/methods/response/chat/ChatUnfurlResponse.java
vegeris: Oh yes, good catch!

slack-api-client/src/main/java/com/slack/api/methods/request/chat/ChatUnfurlRequest.java
vegeris: No strong feelings but I called it `rawMetadata` to be consistent with the naming for the [unfurls](https://github.com/slackapi/java-slack-sdk/blob/e3f79e0e2512d5be5cf1248327b9ca07580e8cf9/slack-api-client/src/main/java/com/slack/api/methods/request/chat/ChatUnfurlRequest.java#L58) param (`unfurls` and `rawUnfurls`)

(Added a javadoc comment!)

slack-api-client/src/main/java/com/slack/api/methods/impl/MethodsClientImpl.java
vegeris: Yup; updated!

json-logs/samples/api/conversations.list.json
```diff
@@ -125,5 +125,6 @@
   "error": "",
   "needed": "",
   "provided": "",
-  "arg": ""
+  "arg": "",
+  "callstack": ""
```
vegeris, line 129: Should the callstack param be left out of these json-logs files? (I don't see it included for any other APIs)

slack-api-client/src/test/java/test_locally/api/methods_admin_api/FieldValidationTest.java
vegeris: Thanks for calling this out (forgot about this); looks like the tests pass if `getWarnings` is also skipped (updated). Looks like most of the tests in the methods_admin_api/FieldValidationTest.java file also skip `getWarning`

docs/japanese/guides/assistants.md
WilliamBergamin: > Ohh, gotcha; potentially have a separate property for entity+event metadata and check for it in FormBody.Builder toForm(ChatPostMessageRequest req)

Yess I think this might be the way forward

docs/english/guides/ai-apps.md
```diff
@@ -128,7 +128,7 @@ app.blockAction("assistant-generate-numbers", (req, ctx) -> {
         .channel(req.getPayload().getChannel().getId())
         .threadTs(req.getPayload().getMessage().getThreadTs())
         .text("OK, I will generate numbers for you!")
-        .metadata(new Message.Metadata("assistant-generate-numbers", eventPayload))
```
vegeris, line 131: Going to change these back to using the `new` operator to reduce unnecessary changes in this PR

slack-api-model/src/main/java/com/slack/api/model/Attachment.java
vegeris: Removed this property since we're removing it from API responses and events

docs/japanese/guides/assistants.md
vegeris: Updated!

slack-api-client/src/main/java/com/slack/api/methods/RequestFormBuilder.java
```diff
@@ -1600,6 +1604,23 @@ public static FormBody.Builder toForm(ChatPostMessageRequest req) {
...
+                String json = GSON.toJson(metadata);
+                form.add("metadata", json);
+            } else {
+                EntityMetadata[] entities = metadata.getEntities();
+                entities = prepareEntities(metadata.getEntities());
+                metadata.setEntities(entities);
+                String json = GSON.toJson(metadata);
+                form.add("metadata", json);
+            }
+        }
```
zimeg: 🪬 suggestion: It might be nice to include a warning if multiple options are provided for `metadata` since event and entity metadata might overwrite metadata set otherwise? Perhaps similar:

https://github.com/slackapi/java-slack-sdk/blob/b803ad17aee810afd3d2bb6c559db66303e541cd/slack-api-client/src/main/java/com/slack/api/methods/RequestFormBuilder.java#L1520-L1522

slack-api-client/src/main/java/com/slack/api/methods/RequestFormBuilder.java
```diff
@@ -2032,6 +2063,29 @@ public static FormBody.Builder toForm(EmojiListRequest req) {
...
+            metadata = prepareEntityMetadata(metadata);
+            String json = GSON.toJson(metadata, EntityMetadata.class);
+            setIfNotNull("metadata", json, form);
+        }
+
+        setIfNotNull("user_auth_required", req.isUserAuthRequired(), form);
+
+        setIfNotNull("user_auth_url", req.getUserAuthUrl(), form);
+
+        setIfNotNull("error", req.getError(), form);
```
zimeg, line 2106: 🧮 quibble: If this might be moved toward the top of this function?

slack-api-client/src/main/java/com/slack/api/methods/RequestFormBuilder.java
zimeg: 👁️‍🗨️ Edit: I notice similar is remarked in `javadoc` following!

slack-api-client/src/main/java/com/slack/api/methods/RequestFormBuilder.java
```diff
@@ -3124,4 +3183,46 @@ private static <T> String getJsonWithGsonAnonymInnerClassHandling(List<T> tList)
         return GSON_ANONYM_INNER_CLASS_INIT_OUTPUT.equals(json) ? GSON.toJson(new ArrayList<>(tList)) : json;
     }
 
+    private static EntityMetadata[] prepareEntities(EntityMetadata[] entities) {
```
zimeg:
```suggestion
    private static EntityMetadata[] getEntitiesMetadata(EntityMetadata[] entities) {
```

🏷️ suggestion: To make this more clear in calling code I hope!

slack-api-client/src/main/java/com/slack/api/methods/RequestFormBuilder.java
```diff
@@ -3124,4 +3183,46 @@ private static <T> String getJsonWithGsonAnonymInnerClassHandling(List<T> tList)
...
+            entity = prepareEntityMetadata(entity);
+            updatedEntities.add(entity);
+        }
+        EntityMetadata[] entityArray = new EntityMetadata[updatedEntities.size()];
+        entityArray = updatedEntities.toArray(entityArray);
+
+        return entityArray;
+    }
+
+    private static EntityMetadata prepareEntityMetadata(EntityMetadata entity) {
```
zimeg:
```suggestion
    private static EntityMetadata setEntityMetadataFields(EntityMetadata entity) {
```

🏷️ suggestion: Similar thought to avoid "prepare" statements-

slack-api-client/src/main/java/com/slack/api/methods/RequestFormBuilder.java
```diff
@@ -3124,4 +3183,46 @@ private static <T> String getJsonWithGsonAnonymInnerClassHandling(List<T> tList)
...
+        } else if (entity.getEntityPayload().getContentItemFields() != null) {
+            String json = GSON.toJson(entity.getEntityPayload().getContentItemFields(),
+                    EntityPayload.ContentItemFields.class);
+            JsonElement fields = GSON.fromJson(json, JsonElement.class);
+            entity.getEntityPayload().setFields(fields);
+            entity.getEntityPayload().setContentItemFields(null);
+        }
+
+        return entity;
+    }
```
zimeg: 👾 suggestion(non-blocking): I'm not familiar enough in `java` to know if this is possible or a good idea, but could we export this from the models implementation file?

```
slack-api-model/src/main/java/com/slack/api/model/EntityMetadata.java
```

slack-api-model/src/main/java/com/slack/api/model/Message.java
```diff
@@ -235,10 +235,24 @@ public void setPublic(boolean isPublic) {
...
+    @NoArgsConstructor
+    @AllArgsConstructor
+    public static class EventAndEntityMetadata {
+        // Event metadata
+        private String eventType;
+        private Map<String, Object> eventPayload;
+
+        // Work object metadata
+        private EntityMetadata[] entities;
+    }
```
zimeg, line 254: 🤖 question: For my own understanding, is this just used as an export? IIRC `model` might reflect responses from the API and I'm unsure how `entities` might be parsed for `metadata` in various cases at a glance...

slack-api-client/src/test/java/test_with_remote_apis/methods/chat_Test.java
zimeg: 💾 @vegeris This is so curious! I wonder what might've changed this file indentation from the others... At a glance we seem to prefer `8` here?

For now let's continue with this change and a note to introduce a shared formatter in CI checks-

slack-api-client/src/test/java/test_with_remote_apis/methods/chat_Test.java
```diff
@@ -44,1056 +48,1329 @@
...
+                                is("https://myappdomain.com/admin/slack/workobject/94/change/"));
+
+                // Verify if the message can be parsed by the JSON parser
+                String ts = postResponse.getTs();
+                ConversationsHistoryResponse history = slack.methods(botToken).conversationsHistory(r -> r
+                                .channel(randomChannelId)
+                                .oldest(ts)
+                                .inclusive(true));
+                assertThat(history.getError(), is(nullValue()));
+        }
```
zimeg: ⭐ praise: Incredible additions!

🌠 ramble: I fear rate limit errors we might find with `conversations.history` but perhaps we soon run these tests with permissive settings?

slack-api-model/src/main/java/com/slack/api/model/Message.java
zimeg: 🧪 edit: I now find the "remote" tests do excellent at covering the cases I was worried about! So nice!

zimeg (approved):
@vegeris Wow! What amazing set of changes this is! 🧰 ✨ 

All recent updates LGTM and I have a note to showcase example code in adjacent places soon, but I did leave a few comments of questions I had while in review.

Nothing seems blocking AFAICT so please feel free to merge when the time is right! 🏁

slack-api-client/src/test/java/test_with_remote_apis/methods/chat_Test.java
vegeris: > rate limit errors we might find with conversations.history

Ah [true](https://docs.slack.dev/changelog/2025/05/29/rate-limit-changes-for-non-marketplace-apps/); takes effect March 2026, eh :thinking: 

Looks like 6 tests in this file call it as part of verification (ie. 2 existing tests plus these work objects tests). Could be a problem if one tries to run all the integration tests at once

I can either remove this verification, or the guidance can be to run one test per minute 🤡

slack-api-client/src/test/java/test_with_remote_apis/methods/chat_Test.java
vegeris: Ah, I tried installing the VSCode Java support extension from Oracle instead and the formatter uses 4 spaces for tab like the other files (not sure why the Red Hat formatter indents differently for this file specifically) 😛 Going to undo the extra indent in this file

zimeg: Sounds good! The OpenJDK has been useful for me as well. And perhaps we soon bring settings from the sample app for linting here?

```
$ mvn spotless:apply
$ mvn spotless:check
```

🔗 https://github.com/slack-samples/bolt-java-starter-template/blob/7a3952833557ae0cecaca97508a08857838e9cba/pom.xml#L24-L62

slack-api-client/src/test/java/test_with_remote_apis/methods/chat_Test.java
zimeg: Hm I have to become familiar with this timing! I find confusion in both statements:

> New non-Marketplace applications and new installations of existing non-Marketplace applications will be subject to the updated rate limits immediately, as of May 29, 2025.

> Internal customer-built applications are not impacted by these changes.

Since sometimes I find this rate limit when testing new apps. For now let's keep these tests confident with notes to revisit this if errors arise 🚢 💨

slack-api-client/src/main/java/com/slack/api/methods/RequestFormBuilder.java
vegeris: I updated the naming for these helper methods to hopefully make it clearer what they do 👀

slack-api-client/src/main/java/com/slack/api/methods/RequestFormBuilder.java
vegeris: Just wondering, why would we order these 'set property if not null' statements above the statements for the metadata property?

zimeg: 📿 thought: I must have confused adjacent implementations - apologies!

👾 ramble: Was thinking that the properties without additional logic were at the top of methods for quick parsing, but I notice this isn't so true elsewhere:

https://github.com/slackapi/java-slack-sdk/blob/1317675b64bd0a92c7a560e33264a5d8f08b7f9c/slack-api-client/src/main/java/com/slack/api/methods/RequestFormBuilder.java#L1664-L1706

📚 note: Matching the ordering of the docs like you've done is nice! I think this also makes finding attributes still needing support more reasonable in these methods too-

slack-api-client/src/main/java/com/slack/api/methods/RequestFormBuilder.java
vegeris: Ah that warning would be helpful here as well; added some warnings!

slack-api-client/src/test/java/test_with_remote_apis/methods/chat_Test.java
vegeris: I removed the call to conversations.history for the chat.postMessage tests since they validate the output of the method by checking the attachments

slack-api-client/src/main/java/com/slack/api/methods/RequestFormBuilder.java
vegeris: I'm leaning toward keeping it in this file as it's only used in building the request

slack-api-client/src/main/java/com/slack/api/methods/RequestFormBuilder.java
vegeris: Updated the naming here as well

PR #1545, opened 2025-12-22
feat: chat stream helper

srtaalej:
This PR adds a helper to the `chat streaming` methods

```java
ChatStreamHelper stream = (req -> req.
    .channel("C0123456789")
    .threadTs("1700000001.123456")
    .recipientTeamId("T0123456789")
    .recipientUserId("U0123456789")
    .bufferSize(100));
stream.append("**hello wo");
stream.append("rld!**");
ChatStopStreamResponse response = stream.stop();
```

## Requirements

### DIFF
--- a/slack-api-client/src/main/java/com/slack/api/methods/AsyncMethodsClient.java
+++ b/slack-api-client/src/main/java/com/slack/api/methods/AsyncMethodsClient.java
@@ -1016,7 +1016,9 @@ public interface AsyncMethodsClient {
     CompletableFuture<ChatStopStreamResponse> chatStopStream(ChatStopStreamRequest req);
 
     CompletableFuture<ChatStopStreamResponse> chatStopStream(RequestConfigurator<ChatStopStreamRequest.ChatStopStreamRequestBuilder> req);
-
+    
+    AsyncChatStreamHelper asyncChatStreamHelper(RequestConfigurator<AsyncChatStreamHelper.AsyncChatStreamHelperBuilder> req);
+    
     CompletableFuture<ChatUpdateResponse> chatUpdate(ChatUpdateRequest req);
 
     CompletableFuture<ChatUpdateResponse> chatUpdate(RequestConfigurator<ChatUpdateRequest.ChatUpdateRequestBuilder> req);
--- a/slack-api-client/src/main/java/com/slack/api/methods/AsyncMethodsClient.java
+++ b/slack-api-client/src/main/java/com/slack/api/methods/AsyncMethodsClient.java
@@ -1017,7 +1017,7 @@ public interface AsyncMethodsClient {
 
     CompletableFuture<ChatStopStreamResponse> chatStopStream(RequestConfigurator<ChatStopStreamRequest.ChatStopStreamRequestBuilder> req);
     
-    AsyncChatStreamHelper asyncChatStreamHelper(RequestConfigurator<AsyncChatStreamHelper.AsyncChatStreamHelperBuilder> req);
+    AsyncChatStreamHelper chatStream(RequestConfigurator<AsyncChatStreamHelper.AsyncChatStreamHelperBuilder> req);
     
     CompletableFuture<ChatUpdateResponse> chatUpdate(ChatUpdateRequest req);
 
--- a/slack-api-client/src/main/java/com/slack/api/methods/ChatStreamHelper.java
+++ b/slack-api-client/src/main/java/com/slack/api/methods/ChatStreamHelper.java
@@ -47,7 +47,7 @@ public class ChatStreamHelper {
     /**
      * The state of the chat stream.
      */
-    public enum State {
+    private enum StreamState {
         STARTING,
         IN_PROGRESS,
         COMPLETED
@@ -72,7 +72,7 @@ public class ChatStreamHelper {
     @Builder.Default
     private StringBuilder buffer = new StringBuilder();
     @Builder.Default
-    private State state = State.STARTING;
+    private StreamState state = StreamState.STARTING;
     private String streamTs;
 
     /**
@@ -88,7 +88,7 @@ public class ChatStreamHelper {
      * @throws SlackApiException        if a Slack API error occurs
      */
     public ChatAppendStreamResponse append(String markdownText) throws IOException, SlackApiException {
-        if (state == State.COMPLETED) {
+        if (state == StreamState.COMPLETED) {
             throw new SlackChatStreamException("Cannot append to stream: stream state is " + state);
         }
 
@@ -148,7 +148,7 @@ public class ChatStreamHelper {
             List<LayoutBlock> blocks,
             Message.Metadata metadata
     ) throws IOException, SlackApiException {
-        if (state == State.COMPLETED) {
+        if (state == StreamState.COMPLETED) {
             throw new SlackChatStreamException("Cannot stop stream: stream state is " + state);
         }
 
@@ -173,7 +173,7 @@ public class ChatStreamHelper {
             }
 
             streamTs = startResponse.getTs();
-            state = State.IN_PROGRESS;
+            state = StreamState.IN_PROGRESS;
         }
 
         ChatStopStreamResponse response = client.chatStopStream(ChatStopStreamRequest.builder()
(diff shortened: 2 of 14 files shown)


slack-api-client/src/main/java/com/slack/api/methods/ChatStreamHelper.java
```diff
@@ -0,0 +1,252 @@
...
+ * {@code
+ * MethodsClient client = Slack.getInstance().methods(token);
+ * ChatStreamHelper stream = ChatStreamHelper.builder()
+ *     .client(client)
+ *     .channel("C0123456789")
+ *     .threadTs("1700000001.123456")
+ *     .recipientTeamId("T0123456789")
+ *     .recipientUserId("U0123456789")
+ *     .bufferSize(100)
+ *     .build();
```
zimeg: 👁️‍🗨️ thought: This works well but I'm wondering if it's possible to attach the chat stream helper to the `client` for an implementation similar to files upload?

```java
String token = System.getenv("SLACK_BOT_TOKEN");
var client = Slack.getInstance().methods(token);
client.filesUploadV2(req -> req.content("greetings!").channel("C01223456789"));
```

slack-api-client/src/main/java/com/slack/api/methods/ChatStreamHelper.java
```diff
@@ -0,0 +1,252 @@
...
+    private final String recipientUserId;
+
+    /**
+     * The length of markdown_text to buffer in-memory before calling a method.
+     * Increasing this value decreases the number of method calls made for the same amount of text,
+     * which is useful to avoid rate limits.
+     * Default is 100.
+     */
+    @Builder.Default
+    private final int bufferSize = 100;
```
zimeg:
```suggestion
    private final int bufferSize = 256;
```

🧰 suggestion: Let's use base 2 for quick divisions! This'll match defaults of adjacent SDKs too.

slack-api-client/src/main/java/com/slack/api/methods/ChatStreamHelper.java
```diff
@@ -0,0 +1,252 @@
...
+            streamTs = startResponse.getTs();
+            state = State.IN_PROGRESS;
+
+            // Create a response object to return (mimicking the append response structure)
+            response = new ChatAppendStreamResponse();
+            response.setOk(startResponse.isOk());
+            response.setChannel(startResponse.getChannel());
+            response.setTs(startResponse.getTs());
+            response.setWarning(startResponse.getWarning());
+            response.setError(startResponse.getError());
```
zimeg, line 227: ⭐ praise: Nice! I forget if casting between related types might also work here but I think the explicit setting is most clear.

zimeg (commented):
@srtaalej This is awesome to review and test! Thanks for putting it together 👾 ✨ 

I'm leaving a few comments around how we might surface this for the web client that might make it easier to use in calling code, similar to other SDKs, but please let me know if language limits these idea 🙏

slack-api-client/src/main/java/com/slack/api/methods/AsyncMethodsClient.java
```diff
@@ -1016,7 +1016,9 @@ CompletableFuture<AdminConversationsWhitelistListGroupsLinkedToChannelResponse>
     CompletableFuture<ChatStopStreamResponse> chatStopStream(ChatStopStreamRequest req);
 
     CompletableFuture<ChatStopStreamResponse> chatStopStream(RequestConfigurator<ChatStopStreamRequest.ChatStopStreamRequestBuilder> req);
-
+    
+    AsyncChatStreamHelper asyncChatStreamHelper(RequestConfigurator<AsyncChatStreamHelper.AsyncChatStreamHelperBuilder> req);
```
zimeg:
```suggestion
    AsyncChatStreamHelper chatStream(RequestConfigurator<AsyncChatStreamHelper.AsyncChatStreamHelperBuilder> req);
```

🪬 question: Is this something we can surface as `chatStream` to mirror similar methods?

slack-api-client/src/main/java/com/slack/api/methods/ChatStreamHelper.java
```diff
@@ -0,0 +1,250 @@
...
+    private final String channel;
+    private final String threadTs;
+    private final String recipientTeamId;
+    private final String recipientUserId;
+
+    /**
+     * The length of markdown_text to buffer in-memory before calling a method.
+     * Increasing this value decreases the number of method calls made for the same amount of text,
+     * which is useful to avoid rate limits.
+     * Default is 100.
```
zimeg:
```suggestion
     * Default is 256.
```

🔏 todo: Small update to match the variable following!

slack-api-client/src/main/java/com/slack/api/methods/ChatStreamHelper.java
```diff
@@ -0,0 +1,250 @@
...
+     * @throws SlackApiException        if a Slack API error occurs
+     */
+    public ChatStopStreamResponse stop(
+            String markdownText,
+            List<LayoutBlock> blocks,
+            Message.Metadata metadata
+    ) throws IOException, SlackApiException {
+        if (state == State.COMPLETED) {
+            throw new SlackChatStreamException("Cannot stop stream: stream state is " + state);
+        }
```
zimeg, line 153: ⚠️ question: Should the `stop` method throw `SlackChatStreamException` too? I'm not certain so wanted to ask!

slack-api-client/src/main/java/com/slack/api/methods/MethodsClient.java
```diff
@@ -1660,6 +1660,8 @@ ChatScheduleMessageResponse chatScheduleMessage(
 
     ChatStopStreamResponse chatStopStream(RequestConfigurator<ChatStopStreamRequest.ChatStopStreamRequestBuilder> req) throws IOException, SlackApiException;
 
+    ChatStreamHelper chatStreamHelper(RequestConfigurator<ChatStreamHelper.ChatStreamHelperBuilder> req);
```
zimeg:
```suggestion
    ChatStreamHelper chatStream(RequestConfigurator<ChatStreamHelper.ChatStreamHelperBuilder> req);
```

👁️‍🗨️ suggestion: Similar naming to match other SDK implementations!

slack-api-client/src/main/java/com/slack/api/methods/ChatStreamHelper.java
```diff
@@ -0,0 +1,250 @@
...
+public class ChatStreamHelper {
+
+    /**
+     * The state of the chat stream.
+     */
+    public enum State {
+        STARTING,
+        IN_PROGRESS,
+        COMPLETED
+    }
```
zimeg, line 54: 🔭 question(non-blocking): Can we make this private? I don't think this is something that should be changed from outside of this class!

zimeg: 📚 thought(non-blocking): Changing `State` to `StreamState` might be kind in reading later, but also personal preference.

zimeg (commented):
@srtaalej This is testing swell! The one comment I have that we might want to finalize before merging is the naming of this streamer 📸 

The other SDKs have a variation of `chat_stream` or `chatStream` and IMHO we could omit "helper" from calling code, but please let me know if reason exists otherwise. No blocker to the functionalities so I'm in favor of whatever makes most sense!

slack-api-client/src/main/java/com/slack/api/methods/ChatStreamHelper.java
srtaalej: I agree!

slack-api-client/src/main/java/com/slack/api/methods/AsyncMethodsClient.java
srtaalej: yes good idea ⭐

Cr1stal:
Hi @srtaalej and maintainers,

Sorry for opening a duplicate PR that overlapped with this one. I should have checked the existing rules more carefully and coordinated here instead.

If it would be helpful, I’d be happy to help move this PR over the finish line. I can work on addressing review comments, resolving conflicts with `main`, adding or updating tests, or making any follow-up changes the maintainers think are needed.

Thanks, and apologies again for the duplicate work.

PR #1558, opened 2026-02-24
Add @Required annotation with optional validator

fst-john:
Adds a new `@Required` annotation to the model package for use in marking attributes in model classes as "required". This is so we can annotate attributes of our model classes that are expected to be present on every instance of the object, so developers can safely access without any explicit null checks. Meant to be a drop-in replacement for in-line code comments or javadocs that state the same (but apply no actual enforcement at runtime)

* [X] **slack-api-client** (Slack API Clients)
* [X] **slack-api-model** (Slack API Data Models)

## Requirements

### DIFF
--- a/slack-api-model/src/main/java/com/slack/api/util/json/RequiredPropertyDetectionAdapterFactory.java
+++ b/slack-api-model/src/main/java/com/slack/api/util/json/RequiredPropertyDetectionAdapterFactory.java
@@ -7,8 +7,8 @@ import com.google.gson.TypeAdapter;
 import com.google.gson.reflect.TypeToken;
 import com.google.gson.stream.JsonReader;
 import com.google.gson.stream.JsonWriter;
-import com.slack.api.model.predicate.FieldPredicate;
-import com.slack.api.model.annotation.Required;
+import com.slack.api.util.predicate.FieldPredicate;
+import com.slack.api.util.annotation.Required;
 import lombok.EqualsAndHashCode;
 import lombok.RequiredArgsConstructor;
 
@@ -22,7 +22,7 @@ import java.io.IOException;
 /**
  * Adapter factory for processing objects annotated with {@link Required}.  This annotation signals what properties
  * of a model object are required, and thus should be expected to be initialized on instantiated instances. For all
- * fields on the model objected annotated with {@link Required} applies the {@link FieldPredicate#test(Object)} via the
+ * fields on the model objected annotated with {@link Required} applies the {@link FieldPredicate#validate(Object)} via the
  * specified {@link Required#validator()}.
  * <p>
  * Note that this adapter handles both deserialization (JSON --> POJO) and serialization (POJO --> JSON).
--- a/slack-api-model/src/test/java/test_locally/util/JSONUtilityTest.java
+++ b/slack-api-model/src/test/java/test_locally/util/JSONUtilityTest.java
@@ -1,7 +1,6 @@
 package test_locally.util;
 
 import com.google.gson.*;
-import com.slack.api.model.predicate.FieldPredicate;
 import com.slack.api.model.block.ContextBlockElement;
 import com.slack.api.model.block.DividerBlock;
 import com.slack.api.model.block.LayoutBlock;
@@ -12,14 +11,16 @@ import com.slack.api.model.block.element.BlockElement;
 import com.slack.api.model.block.element.ImageElement;
 import com.slack.api.model.block.element.OverflowMenuElement;
 import com.slack.api.model.event.FunctionExecutedEvent;
-import com.slack.api.model.annotation.Required;
+import com.slack.api.util.annotation.Required;
 import com.slack.api.util.json.*;
+import com.slack.api.util.predicate.FieldPredicate;
 import lombok.Builder;
 import lombok.Data;
 import org.junit.Test;
 import test_locally.unit.GsonFactory;
 
 import java.lang.reflect.Type;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
@@ -202,7 +203,7 @@ public class JSONUtilityTest {
     @Test
     public void testRequiredPropertyDetectionAdapterFactory_advancedCase_failureCases() {
         Gson gson = GsonFactory.createSnakeCaseWithRequiredPropertyDetection();
-
+        List<String> users = new ArrayList<>();
         // Serialization
         JsonParseException e = assertThrows(JsonParseException.class, () -> gson.toJson(TestClassWithRequiredAdvanced.builder().build()));
         assertThat(e.getMessage(), equalToIgnoringCase("Required field 'id' failed validation in TestClassWithRequiredAdvanced using predicate IntegerGreaterThanZero"));
@@ -216,7 +217,8 @@ public class JSONUtilityTest {
         e = assertThrows(JsonParseException.class, () -> gson.toJson(TestClassWithRequiredAdvanced.builder().id(1).name("Hello").build()));
         assertThat(e.getMessage(), equalToIgnoringCase("Required field 'users' failed validation in TestClassWithRequiredAdvanced using predicate NonEmptyCollection"));
 
-        e = assertThrows(JsonParseException.class, () -> gson.toJson(TestClassWithRequiredAdvanced.builder().id(1).name("Hello").users(List.of("user1")).build()));
+        users.add("user1");
+        e = assertThrows(JsonParseException.class, () -> gson.toJson(TestClassWithRequiredAdvanced.builder().id(1).name("Hello").users(users).build()));
         assertThat(e.getMessage(), equalToIgnoringCase("Required field 'myBool' failed validation in TestClassWithRequiredAdvanced using predicate isNotNullFieldPredicate"));
 
         // Deserialization
@@ -239,10 +241,12 @@ public class JSONUtilityTest {
     @Test
     public void testRequiredPropertyDetectionAdapterFactory_advancedCase_happyPath() {
         Gson gson = GsonFactory.createSnakeCaseWithRequiredPropertyDetection();
+        List<String> users = new ArrayList<>();
+        users.add("testUser");
         TestClassWithRequiredAdvanced instance = TestClassWithRequiredAdvanced.builder()
                 .id(1)
                 .name("test")
-                .users(List.of("testUser"))
+                .users(users)
                 .myBool(true)
                 .build();
 
--- a/slack-api-model/src/main/java/com/slack/api/util/annotation/Required.java
+++ b/slack-api-model/src/main/java/com/slack/api/util/annotation/Required.java
@@ -1,7 +1,7 @@
-package com.slack.api.model.annotation;
+package com.slack.api.util.annotation;
 
-import com.slack.api.model.predicate.FieldPredicate;
-import com.slack.api.model.predicate.IsNotNullFieldPredicate;
+import com.slack.api.util.predicate.FieldPredicate;
+import com.slack.api.util.predicate.IsNotNullFieldPredicate;
 import com.slack.api.util.json.RequiredPropertyDetectionAdapterFactory;
 
 import java.lang.annotation.ElementType;
--- a/slack-api-model/src/main/java/com/slack/api/util/predicate/FieldPredicate.java
+++ b/slack-api-model/src/main/java/com/slack/api/util/predicate/FieldPredicate.java
@@ -1,8 +1,8 @@
-package com.slack.api.model.predicate;
+package com.slack.api.util.predicate;
 
 /**
  * A functional interface for defining validation predicates against {@link java.lang.reflect.Field}.  Used by
- * {@link com.slack.api.model.annotation.Required} during object serialization and deserialization to ensure the
+ * {@link com.slack.api.util.annotation.Required} during object serialization and deserialization to ensure the
  * field member is "valid" per the defined predicate.
  */
 @FunctionalInterface
--- a/slack-api-model/src/main/java/com/slack/api/util/predicate/IsNotNullFieldPredicate.java
+++ b/slack-api-model/src/main/java/com/slack/api/util/predicate/IsNotNullFieldPredicate.java
@@ -1,4 +1,4 @@
-package com.slack.api.model.predicate;
+package com.slack.api.util.predicate;
 
 import java.util.Objects;
 


slack-api-model/src/test/java/test_locally/unit/GsonFactory.java
```diff
@@ -28,6 +28,10 @@ public static Gson createSnakeCaseWithoutUnknownPropertyDetection(boolean failOn
     }
 
     public static Gson createSnakeCase(boolean failOnUnknownProperties, boolean unknownPropertyDetection) {
+        return getBuilder(failOnUnknownProperties, unknownPropertyDetection).create();
+    }
+
+    public static GsonBuilder getBuilder(boolean failOnUnknownProperties, boolean unknownPropertyDetection) {
```
WilliamBergamin: From my understanding the goal of this factory if to return `Gson` not `GsonBuilder`, what do you think about following the existing pattern and adding something like 

```java
public static Gson createSnakeCase(boolean failOnUnknownProperties, boolean unknownPropertyDetection, requiredPropertyDetection boolean)
....
        if (unknownPropertyDetection) {
            builder.registerTypeAdapterFactory(new  UnknownPropertyDetectionAdapterFactory());
        } 
        if (requiredPropertyDetection) {
            builder.registerTypeAdapterFactory(new  UnknownPropertyDetectionAdapterFactory());
        } 
        return builder.create();
         
```

slack-api-client/src/main/java/com/slack/api/util/json/GsonFactory.java
```diff
@@ -86,5 +88,9 @@ public static void registerTypeAdapters(GsonBuilder builder, boolean failOnUnkno
                 .registerTypeAdapter(AppWorkflow.StepInputValueElementDefault.class, new GsonAppWorkflowStepInputValueDefaultFactory(failOnUnknownProps))
                 .registerTypeAdapter(LogsResponse.DetailsChangedValue.class, new GsonAuditLogsDetailsChangedValueFactory(failOnUnknownProps))
                 .registerTypeAdapter(LogsResponse.UserIDs.class, new GsonAuditLogsDetailsUserIDsFactory(failOnUnknownProps));
+
+        if (failOnRequiredProperties) {
+            builder.registerTypeAdapterFactory(new RequiredAdapterFactory());
+        }
```
WilliamBergamin: This might not be DRY, but what do you think about following the existing pattern and moving this logic inside `createSnakeCase` and `createCamelCase` to follow how `UnknownPropertyDetectionAdapterFactory` is set 🙏

slack-api-client/src/main/java/com/slack/api/SlackConfig.java
```diff
@@ -62,6 +62,9 @@ public void setFailOnUnknownProperties(boolean failOnUnknownProperties) {
             throwException();
         }
 
+        @Override
+        public void setFailOnRequiredProperties(boolean failOnRequiredProperties) { throwException(); }
```
WilliamBergamin, line 66: Nice 💯

slack-api-client/src/main/java/com/slack/api/SlackConfig.java
```diff
@@ -248,6 +251,13 @@ public void setLibraryMaintainerMode(boolean libraryMaintainerMode) {
      */
     private boolean failOnUnknownProperties = false;
 
+    /**
+     * Makes it so that any fields annotated with {@link com.slack.api.model.annotation.Required} that are missing
+     * or invalid when deserializing responses from the Slack Web API client will throw an exception.
+     * By default, this is "off", but can be opted into by setting to true.
```
WilliamBergamin: Nit
```suggestion
     * If you would like to detect required properties by throwing exceptions, set this flag as true.
```

slack-api-model/src/main/java/com/slack/api/model/annotation/FieldPredicate.java
```diff
@@ -0,0 +1,5 @@
+package com.slack.api.model.annotation;
```
WilliamBergamin: I'm thinking we move this to a new package/directory
```json
/**
 * Provides JSON custom annotation utilities for the classes in this library.
 */
package com.slack.api.util.annotations;
```

slack-api-model/src/main/java/com/slack/api/util/json/RequiredPropertyDetectionAdapterFactory.java
```diff
@@ -0,0 +1,111 @@
...
+                    FieldPredicate predicate = annotation.validator().getDeclaredConstructor().newInstance();
+                    entries.add(new RequiredFieldEntry(field, predicate));
+                } catch (NoSuchMethodException | InstantiationException |
+                         IllegalAccessException | InvocationTargetException e) {
+                    throw new JsonParseException(
+                            "Cannot instantiate validator for field: " + field.getName(), e);
+                }
+            }
+        }
+        return Collections.unmodifiableList(entries);
```
WilliamBergamin, line 83: 💯

slack-api-model/src/test/java/test_locally/util/JSONUtilityTest.java
```diff
@@ -153,4 +160,72 @@ public void testGsonFunctionExecutedEventInputValueFactory() {
...
+                return obj instanceof Integer && (int)obj > 0;
+            }
+        }
+
+        public static class NonEmptyString implements FieldPredicate {
+            @Override
+            public boolean test(Object obj) {
+                return obj instanceof String && !((String) obj).isEmpty();
+            }
+        }
```
WilliamBergamin, line 298: This is pretty sweet!

slack-api-model/src/main/java/com/slack/api/model/annotation/FieldPredicate.java
```diff
@@ -0,0 +1,5 @@
...
+
+public interface FieldPredicate {
+    boolean test(Object obj);
```
WilliamBergamin: Nice job on the interface approach 🚀 but I think the `test` terminology might be confusing, what do you think of `check` or `validate` instead?

Maybe `validate` is better since it aligns with `validator`

slack-api-model/src/test/java/test_locally/util/JSONUtilityTest.java
```diff
@@ -153,4 +160,72 @@ public void testGsonFunctionExecutedEventInputValueFactory() {
         parsed = f.deserialize(json, FunctionExecutedEvent.InputValue.class, context);
         assertThat(parsed.asStringArray(), is(Arrays.asList("C111", "C222")));
     }
+
```
WilliamBergamin, line 167: Lets add some test that expect to **_not_** throw an exception, this should improve the coverage for the overrides of `TypeAdapter.write` and `TypeAdapter.read`

Might be a good idea to have those for each json types (array, object, string, number, boolean or null)

slack-api-model/src/main/java/com/slack/api/util/json/RequiredAdapterFactory.java
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Adapter factory for processing objects annotated with {@link Required}.  This annotation signals what properties
+ * of a model object are required, and thus should be expected to be initialized on instantiated instances. For all
+ * fields on the model objected annotated with {@link Required} applies the {@link FieldPredicate#test(Object)} via the
+ * specified {@link Required#validator()}.
+ * <p>
+ * Note that this adapter handles both deserialization (JSON --> POJO) and serialization (POJO --> JSON).
+ */
+public class RequiredAdapterFactory implements TypeAdapterFactory {
```
WilliamBergamin: What do you thinking about following the same naming pattern as `UnknownPropertyDetectionAdapterFactory` and naming this class `RequiredPropertyDetectionAdapterFactory`

slack-api-model/src/test/java/test_locally/unit/GsonFactory.java
fst-john: Yep for the developer-facing factory we want to return the `Gson` instance directly for serialization/deserialization.  This particular factory is just used for the unit test suite, so I amended this slightly so we could have access to the exposed `GstonBuilder` instance directly

slack-api-client/src/main/java/com/slack/api/util/json/GsonFactory.java
fst-john: Hmm yeah I wasn't sure what the intended use of `SlackConfig.isLibraryMaintainerMode`.  Also, most of these `TypeAdapter` that we define don't even rely on `failOnUnknownProps`, for example in `GsonFileFactory` (among others) it's not used at all
https://github.com/slackapi/java-slack-sdk/blob/797d1c1ac9a4a0d209e43c680f1c0ebbd8263e61/slack-api-model/src/main/java/com/slack/api/util/json/GsonFileFactory.java#L15

I think a better pattern would actually be 
```java
public static Gson createSnakeCase() {
   GsonBuilder builder = new GsonBuilder().setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES);
   configureGsonBuilder(builder);
   return builder.build();
}

public static Gson createSnakeCase(SlackConfig config) {
   GsonBuilder builder = new GsonBuilder().setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES);
  configureGsonBuilder(builder, config);
  return builder.build();
}

....

private static void configureGsonBuilder(GsonBuilder builder) {
   configureGsonBuilder(builder, SlackConfig.DEFAULT);
}
private static void configureGsonBuilder(GsonBuilder builder, SlackConfig config) {
   boolean failOnUnknownProps = config.isFailOnUnknownProperties();
   boolean failOnRequiredProps = config.isFailOnRequiredProperties();
   if (failOnUnknownProps || config.isLibraryMaintainerMode()) {
      builder.registerTypeAdapterFactory(new UnknownPropertyDetectionAdapaterFactory());
   }
  if (failOnRequiredProps) {
    builder.registerTypeAdapaterFactory(new RequiredAdapaterFactory());
  }

  ... // Set other TypeAdapters that we register for slack content
}
```

WDYT?

slack-api-model/src/main/java/com/slack/api/model/annotation/FieldPredicate.java
fst-john: Yeah that's fair; I went with `test` because that's the abstract method on the `Predicate` functional interface https://docs.oracle.com/javase/8/docs/api/java/util/function/Predicate.html

slack-api-model/src/main/java/com/slack/api/util/json/RequiredAdapterFactory.java
fst-john: Love it - yeah that's a way better name!

slack-api-model/src/main/java/com/slack/api/model/annotation/FieldPredicate.java
fst-john: Given that these predicates are meant to live on slack model objects I think it makes sense for it to live in the `model` package, but I'm open to  a new sub-module under `model`; maybe `predicates` (or something?)

slack-api-model/src/test/java/test_locally/util/JSONUtilityTest.java
fst-john: 🫡

slack-api-client/src/main/java/com/slack/api/util/json/GsonFactory.java
```diff
@@ -30,61 +30,83 @@ private GsonFactory() {
...
         }
-        return gsonBuilder.create();
+        if (isPrettyResponseLoggingEnabled) {
+            builder.setPrettyPrinting();
+        }
+
+        registerTypeAdapters(builder, failOnUnknownProps);
     }
 
     public static void registerTypeAdapters(GsonBuilder builder, boolean failOnUnknownProps) {
```
fst-john, line 82: I don't know why this is `public` but I don't wanna break anybody, so leaving it as is 🤷‍♂️

slack-api-client/src/main/java/com/slack/api/util/json/GsonFactory.java
WilliamBergamin: This refactoring work feels a bit out of scope for this PR, lets add the following to `createSnakeCase` and `createCamelCase` instead, if we want to do this refactor we can open a follow up PR for it 🙏 

```java
 if (failOnRequiredProps) {
      builder.registerTypeAdapterFactory(new RequiredPropertyDetectionAdapterFactory());
}
```


I think `isLibraryMaintainerMode` is used when we generate json logs from the payloads 🤔 there is a separate process for this

slack-api-model/src/test/java/test_locally/unit/GsonFactory.java
WilliamBergamin: Yess this is only used for tests, but it breaks the "pure" factory pattern that was in place by blending it with a builder pattern, if we make `getBuilder` private and update `createSnakeCase` then we can keep the "pure" factory pattern

slack-api-client/src/main/java/com/slack/api/util/json/GsonFactory.java
fst-john: 🫡 as I was writing it I was like hmmm maybe this is too much haha.  Let me revert these changes and I'll do as you suggested

slack-api-client/src/main/java/com/slack/api/util/json/GsonFactory.java
```diff
@@ -44,12 +44,17 @@ public static Gson createSnakeCase(SlackConfig config) {
         GsonBuilder gsonBuilder = new GsonBuilder()
                 .setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES);
         registerTypeAdapters(gsonBuilder, failOnUnknownProps);
+
         if (failOnUnknownProps || config.isLibraryMaintainerMode()) {
-            gsonBuilder = gsonBuilder.registerTypeAdapterFactory(new UnknownPropertyDetectionAdapterFactory());
```
fst-john, line 48: This reassignment is redundant, so removing it

slack-api-client/src/main/java/com/slack/api/util/json/GsonFactory.java
```diff
@@ -44,12 +44,17 @@ public static Gson createSnakeCase(SlackConfig config) {
...
+            gsonBuilder.registerTypeAdapterFactory(new UnknownPropertyDetectionAdapterFactory());
+        }
+        if (config.isFailOnRequiredProperties()) {
+            gsonBuilder.registerTypeAdapterFactory(new RequiredPropertyDetectionAdapterFactory());
         }
         if (config.isPrettyResponseLoggingEnabled()) {
-            gsonBuilder = gsonBuilder.setPrettyPrinting();
```
fst-john, line 51: Same thing here - there's no need for the re-assignmen

PR #1590, opened 2026-05-15
fix(model): handle array format for ListRecord.Field message deserial…

### LINKED ISSUE #1587: ListRecord.Field.message` deserialization fails when Slack returns an array
## Issue summary

  When the Slack API returns a `conversations.replies` (or `conversations.history`) response that contains a Slack Lists item attachment, the SDK throws `JsonSyntaxException` whenever any `ListRecord.Field.message` value is a JSON
  array instead of a single object. The whole API call aborts and no `Message` data is returned.

  This is the same class of bug as #1450 (open) and #1454 (closed as dup of #1450). Both are still observable against the latest release v1.48.1 — reproduced below with a minimal payload.

  ## Affected versions

  - v1.46.0 through v1.48.1 (verified)
  - Likely earlier versions back to when the field was introduced

  ## Root cause

  `com.slack.api.model.list.ListRecord.Field` declares `message` as a single `Message`:
  
  ```java
  // slack-api-model/src/main/java/com/slack/api/model/list/ListRecord.java
  public static class Field {
      ...
      private Message message;
      ...
  }
  ```

  Slack's API can return this value as an array of `Message` objects, and Gson then throws:

  ```
  com.google.gson.JsonSyntaxException: java.lang.IllegalStateException:
    Expected BEGIN_OBJECT but was BEGIN_ARRAY at line 1 column N path
    $.messages[I].attachments[J].list_record.record.fields[K].message
  ```

  Every other field on `ListRecord.Field` is modeled as a `List<X>` (`number`, `select`, `date`, `user`, `attachment`, `email`, `phone`, `channel`, `rating`, `timestamp`, `link`, `reference`) — only `message` is a scalar. That
  asymmetry is what makes this fail.

  ## Minimal reproduction

  ```java
  import com.google.gson.Gson;
  import com.slack.api.methods.response.conversations.ConversationsRepliesResponse;
  import com.slack.api.util.json.GsonFactory;

  public class Repro {
      public static void main(String[] args) {
          // Mirrors the shape Slack returns when a Lists item's `message` field
          // has multiple values (array form).
          String payload =
              "{\"ok\":true,\"messages\":[{\"type\":\"message\",\"user\":\"U1\",\"ts\":\"1.0\","
              + "\"text\":\"hi\",\"attachments\":[{\"list_record\":{\"record\":{\"fields\":["
              + "{\"key\":\"name\",\"message\":{\"text\":\"ok\"}},"
              + "{\"key\":\"details\",\"message\":[{\"text\":\"a\"},{\"text\":\"b\"}]}"
              + "]}}}]}]}";

          Gson sdkGson = GsonFactory.createSnakeCase();
          // Throws JsonSyntaxException at
          // $.messages[0].attachments[0].list_record.record.fields[1].message
          ConversationsRepliesResponse parsed =
              sdkGson.fromJson(payload, ConversationsRepliesResponse.class);
      }
  }
  ```

  ## Expected behavior

  The SDK should accept both single-object and array shapes for `ListRecord.Field.message`, matching Slack's actual API behavior.

hello-ashleyintech:
…ization

Slack's API can return the `message` field in ListRecord.Field as eithera JSON object or a JSON array. Previously, receiving an array caused an error `JsonSyntaxException ("Expected BEGIN_OBJECT but was
  BEGIN_ARRAY")`.

Added `GsonListRecordFieldFactory` to normalize both shapes and not cause regressions on existing expected behavior. The existing `getMessage()` API continues to return a single `Message` (first element), and a new getMessages() accessor provides the full list.

Closes #1587

## Requirements

### DIFF
--- a/slack-api-model/src/main/java/com/slack/api/model/list/ListRecord.java
+++ b/slack-api-model/src/main/java/com/slack/api/model/list/ListRecord.java
@@ -11,6 +11,7 @@ import lombok.EqualsAndHashCode;
 import lombok.NoArgsConstructor;
 import lombok.experimental.SuperBuilder;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
@@ -46,6 +47,7 @@ public class ListRecord {
         private String text;
         @SerializedName("rich_text")
         private List<RichTextBlock> richText;
+        private transient List<Message> messages;
         private Message message;
         private List<Double> number;
         private List<String> select;
@@ -60,6 +62,20 @@ public class ListRecord {
         private List<Integer> timestamp;
         private List<LinkField> link;
         private List<ReferenceField> reference;
+
+        public List<Message> getMessages() {
+            if (messages == null && message != null) {
+                return Collections.singletonList(message);
+            }
+            return messages;
+        }
+  
+          public void setMessages(List<Message> messages) {
+            this.messages = messages;
+            if (messages != null && !messages.isEmpty()) {
+                this.message = messages.get(0);
+            }
+        }
     }
 
     /**
--- a/slack-api-model/src/test/java/test_locally/api/model/list/ListRecordFieldTest.java
+++ b/slack-api-model/src/test/java/test_locally/api/model/list/ListRecordFieldTest.java
@@ -1,0 +1,116 @@
+package test_locally.api.model.list;
+
+import com.google.gson.Gson;
+import com.slack.api.model.Message;
+import com.slack.api.model.list.ListRecord;
+import org.junit.Test;
+import test_locally.unit.GsonFactory;
+
+import java.util.List;
+
+import static org.hamcrest.CoreMatchers.*;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class ListRecordFieldTest {
+
+    private final Gson gson = GsonFactory.createSnakeCase();
+
+    @Test
+    public void messageAsObject() {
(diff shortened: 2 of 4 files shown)


slack-api-model/src/main/java/com/slack/api/model/list/ListRecord.java
```diff
@@ -60,6 +62,20 @@ public static class Field {
         private List<Integer> timestamp;
         private List<LinkField> link;
         private List<ReferenceField> reference;
+
+        public List<Message> getMessages() {
+            if (messages == null && message != null) {
+                return Collections.singletonList(message);
```
WilliamBergamin, line 68: Sweeet 🍬 I like this 💯

slack-api-model/src/main/java/com/slack/api/model/list/ListRecord.java
```diff
@@ -60,6 +62,20 @@ public static class Field {
...
+            }
+            return messages;
+        }
+  
+          public void setMessages(List<Message> messages) {
+            this.messages = messages;
+            if (messages != null && !messages.isEmpty()) {
+                this.message = messages.get(0);
```
WilliamBergamin, line 76: I'm not sure of this, but I think in most cases users will know to use `getMessages` if they expect multiple messages 🤔

WilliamBergamin (approved):
Nice work 🚀  this looks good to me, thanks for including unit tests

PR #1611, opened 2026-06-26
feat(model): add table Block Kit block

zimeg:
## Summary

Adds support for the `table` Block Kit layout block to the Java SDK, bringing it to parity with the JS (`@slack/types`) and Python (`slack_sdk`) SDKs, which already ship this block.

A `table` block renders tabular data as a grid of rows and cells:

- **`TableBlock`** (`type: "table"`) — `rows` (list of rows, each a list of cells; up to 100 rows, 20 cells/row), optional `columnSettings`, and `blockId`.
- **`TableCell`** interface implemented by the three cell types:
  - `RawTextTableCell` (`raw_text`) — `text`
  - `RawNumberTableCell` (`raw_number`) — `value`, optional display `text`
  - `RichTextBlock` (`rich_text`) — the existing rich text block now also implements `TableCell`
- **`TableColumnSetting`** — `align` (`left`/`center`/`right`), `isWrapped`
- **`GsonTableCellFactory`** — polymorphic (de)serialization of cells by `type`, registered in `GsonFactory` (and the test-local factory); `TableBlock` registered in `GsonLayoutBlockFactory`
- **`Blocks.table(...)`** builder helper, mirroring the other block helpers

Follows the existing block-model conventions (Lombok `@Data @Builder @NoArgsConstructor @AllArgsConstructor`, `public static final String TYPE`, javadoc linking to the docs).

## Docs

https://docs.slack.dev/reference/block-kit/blocks/table-block

## Validation

Run in this branch's worktree:

- `./mvnw -pl slack-api-model -am test -Dtest=BlockKitTest -Djacoco.skip=true` → **Tests run: 45, Failures: 0, Errors: 0** (includes the 2 new `parseTableBlock` / `buildTableBlock` cases)
- `./mvnw -pl slack-api-client -am compile -Djacoco.skip=true` → **BUILD SUCCESS** (validates the `GsonFactory` registration)

Note: `jacoco.skip=true` was required because the locally available JDK is Java 25 and the project's JaCoCo agent cannot instrument class file major version 69 (`Unsupported class file major version 69`). This is an environment limitation unrelated to these changes; compilation and the tests themselves pass. CI (running the project's supported JDK) should exercise coverage normally.

No changeset added — this repo uses Maven + `jSlack_changelog.md`, not the `.changeset/` workflow.


kevin-boban-glean:
@zimeg Thanks for this PR. Can this be merged please to bring parity?

slack-api-model/src/main/java/com/slack/api/model/block/RawNumberTableCell.java
```diff
@@ -0,0 +1,23 @@
...
+ * A {@code raw_number} table cell, holding a numeric value. The optional {@code text}
+ * field carries the display representation of the value (for example a formatted string).
+ *
+ * @see <a href="https://docs.slack.dev/reference/block-kit/blocks/table-block">Table block</a>
+ */
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class RawNumberTableCell implements TableCell {
```
zimeg, line 18: 👁️‍🗨️ question: Is this pattern more standard than perhaps reversing the order for more "direct" imports?

```diff
- RawNumberTableCell
+ TableCellRawNumber
```

🪬 note: I might prefer the latter but not a blocker!

slack-api-model/src/main/java/com/slack/api/util/json/GsonTableCellFactory.java
```diff
@@ -0,0 +1,56 @@
...
+    public JsonElement serialize(TableCell src, Type typeOfSrc, JsonSerializationContext context) {
+        return context.serialize(src);
+    }
+
+    private Class<? extends TableCell> getTableCellClassInstance(String typeName) {
+        switch (typeName) {
+            case RawTextTableCell.TYPE:
+                return RawTextTableCell.class;
+            case RawNumberTableCell.TYPE:
+                return RawNumberTableCell.class;
```
zimeg, line 49: 🧮 question: Without testing on a personal sandbox I forget if this is stable? It's documented so no blocker here either!

slack-api-model/src/main/java/com/slack/api/model/block/RawNumberTableCell.java
```diff
@@ -0,0 +1,23 @@
...
+    public static final String TYPE = "raw_number";
+    private final String type = TYPE;
+    private Double value;
+    private String text;
+}
```
zimeg, line 23: 📝 note: I find `raw_number` is also supported with the [Data Table](https://docs.slack.dev/reference/block-kit/blocks/data-table-block/#schema-raw-number) block which makes me curious if this should be exported as a composition object instead?

🔗 https://github.com/slackapi/java-slack-sdk/tree/main/slack-api-model/src/main/java/com/slack/api/model/block/composition

zimeg (commented):
📝 Leaving some thoughts on the first shot implementation that I think we might want to consider before merge. I'm curious if we start with a less complete implementation - `table` but no cells - for sake of aligning the[ `data_table`](https://docs.slack.dev/reference/block-kit/blocks/data-table-block/) block properties but I haven't investigated too much more.

PR #1613, opened 2026-06-26
feat(block-kit): add data_table layout block

zimeg:
## Summary

Adds the new `data_table` Block Kit layout block to the Java SDK model. The block displays rich tables that support pagination, sorting, filtering, and interactivity, where each row is a list of cells (`raw_text`, `raw_number`, or `rich_text`).

What's included:
- `DataTableBlock` (type `data_table`) with `rows`, `caption`, `pageSize`, `rowHeaderColumnIndex`, and `blockId`.
- The shared **`TableCell`** cell model in `com.slack.api.model.block.composition`: `RawTextTableCell` (`raw_text`), `RawNumberTableCell` (`raw_number`), and `RichTextBlock` (`rich_text`, which now implements `TableCell`). `DataTableBlock.rows` is `List<List<TableCell>>`.
- `GsonTableCellFactory` to deserialize cells by `type`, registered for `TableCell` in both the client `GsonFactory` and the model test factory.
- `DataTableBlock` registered in `GsonLayoutBlockFactory`.
- `Blocks.dataTable(...)` builder DSL entry.

Field javadoc mirrors the [data-table-block reference](https://docs.slack.dev/reference/block-kit/blocks/data-table-block) Fields table verbatim, kept in sync with the corresponding node SDK PR ([slackapi/node-slack-sdk#2638](https://github.com/slackapi/node-slack-sdk/pull/2638)) after its review.

### Shared cell model with the `table` block

The `data_table` cells are the same `raw_text` / `raw_number` / `rich_text` cell model the `table` block uses, so this PR uses the shared **`TableCell`** family (interface + `RawTextTableCell` / `RawNumberTableCell` in `...block.composition`, deserialized by a single `GsonTableCellFactory`) rather than a parallel `data_table`-only set. This is the same cell model the sibling **`table`-block PR #1611** introduces — identical definitions, same package — so the two are aligned by construction: **whichever lands second drops the now-duplicate cell files in a trivial merge.** No behavioral reconciliation needed. (Mirrors node and python, which likewise share one cell model across both blocks — node's `@slack/types` composition objects, python's `slack_sdk.models.blocks.basic_components`.)

## Reference

Schema: https://docs.slack.dev/reference/block-kit/blocks/data-table-block

## Validation

Run in the touched modules (against latest `main` merged in):

- `./mvnw -pl slack-api-model -am test -Dtest=BlockKitTest -Djacoco.skip=true` — Tests run: 57, Failures: 0, Errors: 0. Includes `parseDataTableBlock` (deserialization of all three cell types + `page_size`/`row_header_column_index`) and `buildDataTableBlock` (builder DSL + round-trip serialization).
- `./mvnw -pl slack-api-client -am test-compile -Djacoco.skip=true` — BUILD SUCCESS (verifies the `GsonFactory` change in the client module compiles).

Note: `jacoco.skip=true` was required because the locally available JDK is 25 and the pinned JaCoCo agent (0.8.7) cannot instrument class file major version 69. This is an environment limitation unrelated to the change; the surefire test execution itself passes. CI runs the JDK 8/14/17 matrix, where JaCoCo is unaffected.

This repo does not use changesets (Maven multi-module project), so no changeset was added.

Co-Authored-By: Claude <svc-devxp-claude@slack-corp.com>


slack-api-model/src/main/java/com/slack/api/model/block/composition/TableCell.java
zimeg, line 1: 🦠 note: I have some confidence this is a consistent shared type but it's not clear if it's meant to be stable because it's not in reference. Hoping to add this here and https://github.com/slackapi/java-slack-sdk/pull/1611 for shared implementation!

slack-api-client/src/main/java/com/slack/api/util/json/GsonFactory.java
```diff
@@ -80,6 +81,7 @@ public static void registerTypeAdapters(GsonBuilder builder, boolean failOnUnkno
                 .registerTypeAdapter(ContextActionsBlockElement.class, new GsonContextActionsBlockElementFactory(failOnUnknownProps))
                 .registerTypeAdapter(BlockElement.class, new GsonBlockElementFactory(failOnUnknownProps))
                 .registerTypeAdapter(RichTextElement.class, new GsonRichTextElementFactory(failOnUnknownProps))
+                .registerTypeAdapter(TableCell.class, new GsonTableCellFactory(failOnUnknownProps))
```
zimeg, line 84: 📢 note: Here we add custom deserialization to handle the various `type` possibilities of the table cell.

slack-api-model/src/main/java/com/slack/api/model/block/composition/BlockCompositions.java
```diff
@@ -54,6 +54,26 @@ public static MarkdownTextObject markdownText(String text) {
...
 
+    // RawNumberObject
+
+    public static RawNumberObject rawNumber(ModelConfigurator<RawNumberObject.RawNumberObjectBuilder> configurator) {
+        return configurator.configure(RawNumberObject.builder()).build();
+    }
+
+    public static RawNumberObject rawNumber(Double value, String text) {
+        return RawNumberObject.builder().value(value).text(text).build();
+    }
```
zimeg, line 65: 👁️‍🗨️ note: These introduce helper functions to `BlockComposition` that aren't documented but IMHO are helpful for:

```diff
- RawTextObject.builder().text("Name").build()
+ BlockCompositions.rawText("Name"),
```

zimeg (commented):
☕ Ideas to share where included I callout a pattern I'm uncertain of around `rawText` and `rawNumber`!

PR #1615, opened 2026-06-29
docs: clarify thread_ts is optional for assistant.threads.setSuggestedPrompts

zimeg:
This pull request clarifies that the `thread_ts` argument of `assistant.threads.setSuggestedPrompts` is optional.

- `thread_ts` is no longer required by the API. In the Messages tab agent experience, suggested prompts live at the top of the Messages tab rather than within a thread, so a `thread_ts` is not needed.
- `AssistantThreadsSetSuggestedPromptsRequest.threadTs` is a Lombok builder field already sent via `setIfNotNull`, so it was already optional at runtime — this updates the Javadoc to say so.
- Removes a stale copy-paste reference to "where to set the status" (carried over from `setStatus`).

### Category

## Requirements

I've read the [Contributing guidelines](https://github.com/slackapi/java-slack-sdk/blob/main/.github/contributing.md) and [Code of Conduct](https://slackhq.github.io/code-of-conduct).

### DIFF
--- a/slack-api-client/src/main/java/com/slack/api/methods/request/assistant/threads/AssistantThreadsSetSuggestedPromptsRequest.java
+++ b/slack-api-client/src/main/java/com/slack/api/methods/request/assistant/threads/AssistantThreadsSetSuggestedPromptsRequest.java
@@ -24,7 +24,7 @@ public class AssistantThreadsSetSuggestedPromptsRequest implements SlackApiReque
     private String channelId;
 
     /**
-     * Message timestamp of the thread of where to set the status.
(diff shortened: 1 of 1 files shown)


zimeg:
Closing: `thread_ts` is already optional in the Java SDK at runtime — the builder field has no `@NonNull` and `RequestFormBuilder` sends it via `setIfNotNull`, so calling `setSuggestedPrompts` without a `thread_ts` already produces the correct request. No code change is needed; only the Javadoc was slightly stale, which isn't worth a standalone PR.

PR #1622, opened 2026-07-14
feat(methods): add blocks.validate method

zimeg:
This pull request adds the `blocks.validate` Web API method to the Slack API client for Java.

- Callers validate Block Kit payloads via `slack.methods().blocksValidate(req)` and the async equivalent `slack.methodsAsync().blocksValidate(req)`.
- `blocks.validate` is **unauthenticated** — no token or scopes required — so the request carries no token (modeled like `api.test`).
- Rate limiting: modeled as a special tier (`SpecialTier_blocks_validate`) per the [method reference](https://docs.slack.dev/reference/methods/blocks.validate).

## Request — typed inputs with string escape hatches

Each of the three payload kinds accepts a typed value **or** a raw JSON string (the `*AsString` field; the string wins if both are set), mirroring `ChatPostMessageRequest` / `ViewsOpenRequest`:

| field | typed | string |
| --- | --- | --- |
| blocks | `List<LayoutBlock>` | `blocksAsString` |
| message | `MessagePayload` | `messageAsString` |
| view | `View` | `viewAsString` |

`MessagePayload` is a small nested `{blocks, attachments}` type rather than the response-shaped `Message` model. The accepted shape was verified against the live endpoint: `attachments` (with nested blocks) are validated, whereas `text` is rejected as an additional property — so `text` is intentionally omitted, and the type is valid-by-construction.

## Response — validation errors

On a failed validation the response returns `errors[]`, each with `pointer` (a JSON pointer to the offending element), `code`, `message`, and a `constraint` object. The constraint's `expected`/`got` are modeled as `JsonElement` because their shape is polymorphic across constraint types (a string array for `enum`, a number for `min_length`/`max_length`/`max_items`, or absent for `all_of`) — a single static type here caused a Gson `JsonSyntaxException` on numeric constraints.

## Testing

- `./mvnw test -pl slack-api-client -Dtest=test_locally.api.methods.BlocksTest` — local mock server, sync + async, typed and string paths (4 tests).
- `test_with_remote_apis.methods.blocks_Test` exercises the live API: a well-formed payload, a malformed payload (`invalid_blocks` + populated `errors[]`), and a numeric-constraint case (`max_items`) that guards the polymorphic `expected`/`got` deserialization (3 tests).

### Companion example

An end-to-end example lives in slack-samples/bolt-java-examples#53, building blocks with the SDK builders and validating them.

## Category

## Requirements


slack-api-client/src/main/java/com/slack/api/methods/request/blocks/BlocksValidateRequest.java
```diff
@@ -0,0 +1,60 @@
...
+    public String getToken() {
+        return null;
+    }
+
+    @Data
+    @Builder
+    public static class MessagePayload {
+        private List<LayoutBlock> blocks;
+        private List<Attachment> attachments;
+    }
```
zimeg, line 59: 👁️‍🗨️ note: This is added to the **request** class for `blocks.validate` to bring matching arguments for:

- blocks: blocksAsString
- message: messageAsString
- view: viewsAsString

slack-api-client/src/main/java/com/slack/api/methods/response/blocks/BlocksValidateResponse.java
```diff
@@ -0,0 +1,41 @@
...
+        private String message;
+        private Constraint constraint;
+
+        // expected/got are JsonElement because their shape varies by constraint type
+        // (a string array for "enum", a number for length/count constraints, or absent).
+        @Data
+        public static class Constraint {
+            private String type;
+            private JsonElement expected;
+            private JsonElement got;
```
zimeg, line 38: 📣 note: These can be handled with a typecheck in calling code but for now we avoid parsing responses into particular types here.

zimeg (commented):
📝 Callouts for the kind folks checking this out!

WilliamBergamin (approved):

zimeg:
@WilliamBergamin Thanks for a review of this wonderful method! I'll merge this now.

PR #1644, opened 2026-09-10
test: probe ListRecord.Field message wire shape against live API (experimental)

zimeg:
> [!WARNING]
> **Experimental / not for merge.** This is a throwaway probe used to observe the real production response shape of the Slack Lists `message` field before committing to a fix for #1587. Expected to be reverted.

This pull request adds an experimental remote-API test that empirically observes how the Slack API returns a list item's `message` field.

- Motivation: #1587 reports that `ListRecord.Field.message` can come back as an **array** of message objects, which the production Gson path fails to deserialize (`JsonSyntaxException: Expected BEGIN_OBJECT but was BEGIN_ARRAY`). The public docs for `slackLists.items.create` show only the request shape (an array of permalink URL strings) and give no response example, so the response shape is unverified in public.
- Adds `messageFieldArrayShapeProbe` to `slacklists_Test`: creates a list with a `message`-type column, sets the field with a real message permalink, reads the item back, and:
  1. captures the **raw** `slackLists.items.info` JSON and asserts whether the `message` value is a JSON array (`[`) or object (`{`);
  2. separately probes whether the **typed** (production Gson) deserialization throws on that real shape.
- Additive, test-only. No SDK source changes.

### Category

### Testing

Run only the Slack Lists remote test against a workspace with a bot token that has `lists:read`/`lists:write` and can post to the test channel:

```
mvn test -pl slack-api-client -Dtest="test_with_remote_apis.methods.slacklists_Test#messageFieldArrayShapeProbe" -DfailIfNoTests=false
```

Inspect the `[probe]` log lines — the `SUMMARY` line reports `messageIsArray`, `messageIsObject`, and `typedDeserializationThrew` for the live response.

## Requirements

### DIFF
--- a/slack-api-client/src/test/java/test_with_remote_apis/methods/slacklists_Test.java
+++ b/slack-api-client/src/test/java/test_with_remote_apis/methods/slacklists_Test.java
@@ -1,8 +1,14 @@
 package test_with_remote_apis.methods;
 
 import com.slack.api.Slack;
+import com.slack.api.methods.MethodsClient;
 import com.slack.api.methods.SlackApiException;
 import com.slack.api.methods.response.auth.AuthTestResponse;
+import com.slack.api.methods.response.chat.ChatGetPermalinkResponse;
+import com.slack.api.methods.response.chat.ChatPostMessageResponse;
+import com.slack.api.util.http.SlackHttpClient;
+import okhttp3.FormBody;
+import okhttp3.Response;
 import com.slack.api.methods.response.slack_lists.SlackListsAccessDeleteResponse;
 import com.slack.api.methods.response.slack_lists.SlackListsAccessSetResponse;
 import com.slack.api.methods.response.slack_lists.SlackListsCreateResponse;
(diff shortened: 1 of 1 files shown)


zimeg:
**Ran the probe against a live workspace (single scoped test, no manifest/audit tests).** Empirical result:

```
>>>PROBE>>> messagePresent=true messageIsArray=true messageIsObject=false firstValueChar=[
            typedDeserializationThrew=true
            typedError=JsonSyntaxException: Expected BEGIN_OBJECT but was BEGIN_ARRAY
              at ... path $.record.fields[0].message
```

Two confirmations from the live `slackLists.items.info` response:

1. **The `message` field is returned as a JSON array** (the raw value begins with `[`).
2. **The production Gson path throws** `Expected BEGIN_OBJECT but was BEGIN_ARRAY at $.record.fields[0].message` — i.e. #1587 reproduces against production on the current `api` branch. The array-handling adapter added in #1590 lives only in the test tree and is not registered in the shipped `GsonFactory`, so real deserialization still fails.

This validates the direction: the fix must register the `ListRecord.Field` message adapter (or equivalent array-aware deserialization) in the production `GsonFactory`, and the sample/response typing should be array-capable. This PR itself is throwaway — proof only.

PR #1645, opened 2026-09-10
feat: support mrkdwn descriptions on OptionObject (#1471)

### LINKED ISSUE #1471: Support description of mrkdwn type in OptionObject
Need to read options in a checkbox from Slack using the Java SDK. Currently, description text objects inside an option are always coerced into a PlainTextObject even if their type property is set to `mrkdwn`.

### Requirements
Slack supports options with description text objects that use `mrkdwn` type. The SDK coerces them into a PlainTextObject, resulting in a lossy translation.

Consider changing the description field in the `com.slack.api.model.block.composition.OptionObject` to a `com.slack.api.model.block.composition.TextObject` so it can support either plain text or markdown.

zimeg:
## Summary

`OptionObject.description` was hard-typed to `PlainTextObject`, so a `mrkdwn` description on a radio-button or checkbox option was coerced to plain_text on deserialize — the markdown was lost. The [option-object docs](https://docs.slack.dev/reference/block-kit/composition-objects/option-object) state that radio buttons and checkboxes *can* use `mrkdwn` text objects for the description.

Fixes #1471.

## What changed

- **`OptionObject.description`**: `PlainTextObject` → `TextObject` (the same type already used for `OptionObject.text`). This routes deserialization through the existing polymorphic text-object factory, so a `mrkdwn` description round-trips as a `MarkdownTextObject` and a `plain_text` description still returns a `PlainTextObject`.
- **Kotlin DSL** (`OptionObjectBuilder`): added a `markdownDescription(text, verbatim)` overload alongside the existing plain_text `description(text, emoji)`.
- **Tests** (extended existing cases rather than adding standalone ones):
  - `BlockKitTest#parseCheckboxes` — now asserts a `mrkdwn` option description deserializes to `MarkdownTextObject` while a `plain_text` option stays `PlainTextObject`.
  - `ActionsBlockTest#`Channel select and checkboxes`` (Kotlin) — the first checkbox option now serializes via `markdownDescription()` to a `mrkdwn` object, the second keeps the `plain_text` path.

## Compatibility

Source-compatible: existing call sites all pass plain_text, and the getter's declared type only broadens (`PlainTextObject` → its supertype `TextObject`).

## Testing

- `slack-api-model` `BlockKitTest`: 55/55 green
- `slack-api-model-kotlin-extension`: 58/58 green

### DIFF
--- a/slack-api-client/src/test/java/util/sample_json_generation/SampleObjects.java
+++ b/slack-api-client/src/test/java/util/sample_json_generation/SampleObjects.java
@@ -262,10 +262,10 @@ public class SampleObjects {
                     .text(initProperties(PlainTextObject.builder().build()))
                     .build()))
             .options(Arrays.asList(
-                    initProperties(OptionObject.builder().text(initProperties(PlainTextObject.builder().build())).build()),
-                    initProperties(OptionObject.builder().text(initProperties(MarkdownTextObject.builder().build())).build())
+                    initProperties(OptionObject.builder().text(initProperties(PlainTextObject.builder().build())).description(initProperties(PlainTextObject.builder().build())).build()),
+                    initProperties(OptionObject.builder().text(initProperties(MarkdownTextObject.builder().build())).description(initProperties(MarkdownTextObject.builder().build())).build())
             ))
-            .initialOption(initProperties(OptionObject.builder().text(initProperties(PlainTextObject.builder().build())).build()))
+            .initialOption(initProperties(OptionObject.builder().text(initProperties(PlainTextObject.builder().build())).description(initProperties(PlainTextObject.builder().build())).build()))
             .build());
 
     public static List<LayoutBlock> ModalBlocks = asBlocks(
--- a/slack-api-model-kotlin-extension/src/main/kotlin/com/slack/api/model/kotlin_extension/block/composition/OptionObjectBuilder.kt
+++ b/slack-api-model-kotlin-extension/src/main/kotlin/com/slack/api/model/kotlin_extension/block/composition/OptionObjectBuilder.kt
@@ -1,7 +1,9 @@
 package com.slack.api.model.kotlin_extension.block.composition
 
+import com.slack.api.model.block.composition.MarkdownTextObject
 import com.slack.api.model.block.composition.OptionObject
 import com.slack.api.model.block.composition.PlainTextObject
+import com.slack.api.model.block.composition.TextObject
 import com.slack.api.model.kotlin_extension.block.BlockLayoutBuilder
 import com.slack.api.model.kotlin_extension.block.Builder
 import com.slack.api.model.kotlin_extension.block.composition.container.SingleTextObjectContainer
@@ -14,7 +16,7 @@ class OptionObjectBuilder private constructor(
 ) : Builder<OptionObject>, TextObjectDsl by textContainer {
     private var value: String? = null
     private var url: String? = null
-    private var description: PlainTextObject? = null
+    private var description: TextObject? = null
 
     constructor() : this(SingleTextObjectContainer())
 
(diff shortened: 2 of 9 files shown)


slack-api-model/src/main/java/com/slack/api/model/block/composition/OptionObject.java
```diff
@@ -29,11 +29,13 @@ public class OptionObject {
...
-     * below the text field beside the radio button.
-     * Maximum length for the text object within this field is 75 characters.
+     * A plain_text text object that defines a line of descriptive text shown below
+     * the text field beside a single selectable item in a select menu, multi-select
+     * menu, checkbox group, radio button group, or overflow menu. Checkbox group and
+     * radio button group items can also use mrkdwn formatting.
+     * Maximum length for the text within this field is 75 characters.
      */
-    private PlainTextObject description;
+    private TextObject description;
```
zimeg, line 38: 🤓 note: This matches reference and becomes more general to support API requests with validation left to blocks itself.

🔗 https://docs.slack.dev/reference/block-kit/composition-objects/option-object

slack-api-model/src/test/java/test_locally/api/model/block/BlockKitTest.java
zimeg, line 1118: 👽 note: This existed in tests before these change but wasn't checked below.

slack-api-model-kotlin-extension/src/main/kotlin/com/slack/api/model/kotlin_extension/block/composition/OptionObjectBuilder.kt
```diff
@@ -40,15 +42,27 @@ class OptionObjectBuilder private constructor(
...
+    /**
+     * A mrkdwn text object that defines a line of descriptive text shown below the text field beside a single
+     * selectable item. Only checkbox group and radio button group items can use mrkdwn formatting.
+     * Maximum length for the text within this field is 75 characters.
+     *
+     * @see <a href="https://docs.slack.dev/reference/block-kit/composition-objects/option-object">Option object documentation</a>
+     */
+    fun markdownDescription(text: String, verbatim: Boolean? = null) {
+        description = MarkdownTextObject(text, verbatim)
+    }
```
zimeg, line 64: 👁️‍🗨️ note: This pattern is similar to how the text object is supported although I understand this isn't identical for the current description object:

- plainText
- markdownText

🔗 https://github.com/slackapi/java-slack-sdk/blob/dbe498ce0a2f0ed068b4bd7028ce31d22c91998d/slack-api-model-kotlin-extension/src/main/kotlin/com/slack/api/model/kotlin_extension/block/composition/dsl/TextObjectDsl.kt

zimeg: ☁️ note: Calling this out here because I'm uncertain if we're using the best pattern without introducing a break:

- description
- markdownDescription

slack-api-model-kotlin-extension/src/test/kotlin/test_locally/block/ActionsBlockTest.kt
zimeg, line 448: 🧪 note: The existing test suite was modified to check both cases alongside one another.

zimeg (commented):
🗣️ note: A few callouts on these changes that are worth consideration before merge I think!

slack-api-client/src/test/java/util/sample_json_generation/SampleObjects.java
```diff
@@ -262,10 +262,10 @@ private static List<JsonElement> initBlocks() {
                     .text(initProperties(PlainTextObject.builder().build()))
                     .build()))
             .options(Arrays.asList(
-                    initProperties(OptionObject.builder().text(initProperties(PlainTextObject.builder().build())).build()),
-                    initProperties(OptionObject.builder().text(initProperties(MarkdownTextObject.builder().build())).build())
+                    initProperties(OptionObject.builder().text(initProperties(PlainTextObject.builder().build())).description(initProperties(PlainTextObject.builder().build())).build()),
+                    initProperties(OptionObject.builder().text(initProperties(MarkdownTextObject.builder().build())).description(initProperties(MarkdownTextObject.builder().build())).build())
             ))
-            .initialOption(initProperties(OptionObject.builder().text(initProperties(PlainTextObject.builder().build())).build()))
+            .initialOption(initProperties(OptionObject.builder().text(initProperties(PlainTextObject.builder().build())).description(initProperties(PlainTextObject.builder().build())).build()))
```
zimeg, line 268: 🔭 note: These `description` values are added to tests so JSON reflections don't error when deciding between `plain_text` and `mrkdwn` values.

zimeg: 👾 note: This is a test change and wouldn't require similar from developers as I understand.

zimeg (commented):
🪐 A comment of test changes for stable CI.

